PR #5, opened 2018-05-01
Added trainimgs support

avik-pal:
Fixed inconsistency between README and actual source code. Now the
training images of the CIFAR10 Dataset can be directly fetched using the
trainimgs() function.

### DIFF
--- a/src/datasets/cifar10.jl
+++ b/src/datasets/cifar10.jl
@@ -44,7 +44,7 @@ function Base.getindex(v::ValData{BinPackedFS}, i::Integer)
 end
 
 function Base.getindex(v::TrainData{BinPackedFS}, i::Integer)
-    batch, num = divrem((i-1), 10000)
+    batch, num = divrem(i - 1, 10000)
     file = "data_batch_$(batch+1).bin"
     TrainingImage(DataSet, i, open(joinpath(v.set.folder, file)) do f
         seek(f, num * 3073)


src/datasets/autodetect.jl
```diff
@@ -35,14 +35,32 @@ function dataset(which)
...
     elseif which == CIFAR10
         local_path = joinpath(@__DIR__, "..", "..", "datasets", "cifar-10-binary.tar.gz")
-        Base.download("https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz", local_path)
-        run(`tar xzf $local_path`)
+        dir_path = joinpath(@__DIR__,"..","..","datasets","cifar-10-batches-bin")
+        if(!isdir(dir_path))
+            Base.download("https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz", local_path)
+            run(`tar xzf $local_path $dir_path`)
+        end
+        files = ["data_batch_1.bin","data_batch_2.bin","data_batch_3.bin","data_batch_4.bin","data_batch_5.bin"]
```
Keno: I'd rather leave these as is and open the correct file on access to keep the data format compatible with the original formatting.

src/datasets/cifar10.jl
```diff
@@ -41,5 +43,18 @@ function Base.getindex(v::ValData{BinPackedFS}, i::Integer)
     end...)
 end
 
+function Base.getindex(v::TrainData{BinPackedFS}, i::Integer)
+    batch = i ÷ 10000
+    batch = (i % 10000 == 0)? (batch - 1) : batch
+    num = i % 10000
+    num = (num == 0)? (i - 10000 * batch) : num
```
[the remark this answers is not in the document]
avik-pal: It is actually
```julia
batch, num = divrem((i - 1), 10000)
num = num + 1
```
I can make this change. It looks much cleaner

src/datasets/cifar10.jl
```diff
@@ -44,7 +44,7 @@ function Base.getindex(v::ValData{BinPackedFS}, i::Integer)
 end
 
 function Base.getindex(v::TrainData{BinPackedFS}, i::Integer)
-    batch, num = divrem(i-1,10000)
+    batch, num = divrem((i-1), 10000)
```
Keno: No need to parens here

PR #6, opened 2018-05-23
Models imported using ONNX added

ayush1999:
I've added three models (in this PR):
1. Squeezenet
2. DenseNet121
3. ResNet50

All of them are working fine. Binaries could be added after this PR is merged. (along with the new release maybe)


ayush1999:
@MikeInnes Should I add the binaries now, or after merging this? (Because the Travis CI would fail without adding them)

MikeInnes:
Go ahead and add the binaries -- let me know if you need access to do so.

For some of these it would be nice to write these out in a more compact way, e.g. that generates all the layers. This is fine for now though.

ayush1999:
@MikeInnes I might need push access to add the binaries. Also, I was thinking of releasing a new version along with the binaries, if it's not a problem.

PR #9, opened 2018-06-11
Fixed tests and added test for GoogleNet

avik-pal:


avik-pal:
Best to add the tests with the updated API

PR #14, opened 2018-07-05
[WIP] Improve Accuracy of Models

### LINKED ISSUE #3: Pretrained VGG is very bad
It gets most things quite wrong. On slack it was suggested that we may have to look at the way we're doing preprocessing to see if we got something wrong there.

avik-pal:
Fixes #3 
The weights imported for Convolutions are in the form of Cross Correlations. However, Flux performs convolutions and not cross-correlation. So the weights must be flipped about the vertical and horizontal axis before storing them.
I have verified the model by passing random Imagenet Images from both Train and Validation Sets.
CC @staticfloat @ayush1999


avik-pal:
The squeezenet and vgg are working as they are supposed to. But the other models need some more fix (not sure whats wrong though)

src/densenet.jl
```diff
@@ -56,22 +56,22 @@ function densenet_layers()
...
   ls[2].β.data .= weights["conv1/bn_b_0"]
   ls[2].γ.data .= weights["conv1/bn_w_0"]
   l = 4
   for (c, n) in enumerate([6, 12, 24, 16])
       for i in 1:n
           for j in [2, 4]
-              ls[l][i].layer[j].weight.data .= weights["conv$(c+1)_$i/x$(j÷2)_w_0"]
+              ls[l][i].layer[j].weight.data .= weights["conv$(c+1)_$i/x$(j÷2)_w_0"][end:-1:1,:,:,:][:,end:-1:1,:,:]
               ls[l][i].layer[j-1].β.data .= weights["conv$(c+1)_$i/x$(j÷2)/bn_w_0"]
               ls[l][i].layer[j-1].γ.data .= weights["conv$(c+1)_$i/x$(j÷2)/bn_w_0"]
```
staticfloat: It looks to me like we're taking the same weight (`weights["conv$(c+1)_$i/x$(j÷2)/bn_w_0"]`) and assigning it to both beta and gamma here.  This looks like a copy-paste error.

src/densenet.jl
```diff
@@ -56,22 +56,22 @@ function densenet_layers()
...
           end
       end
       l += 2
   end
   for i in [5, 7, 9] # Transition Block Conv Layers
-    ls[i][2].weight.data .= weights["conv$(i÷2)_blk_w_0"]
+    ls[i][2].weight.data .= weights["conv$(i÷2)_blk_w_0"][end:-1:1,:,:,:][:,end:-1:1,:,:]
     ls[i][1].β.data .= weights["conv$(i÷2)_blk/bn_w_0"]
     ls[i][1].γ.data .= weights["conv$(i÷2)_blk/bn_w_0"]
```
staticfloat: Same here as above.

src/densenet.jl
avik-pal: Yes its a copy-paste error. Fixing it

avik-pal:
Still the problem (for densenet) persists. For GoogleNet low accuracy is due to absence of LRN which can only be fixed after we have it in NNlib or Flux.

avik-pal:
@ayush1999 Any reason why the `bias` and `gamma` in case of ResNet was not loaded.             https://github.com/ayush1999/Metalhead.jl/blob/7ca20a54ba98be4f9e6d2c761427422fd7889b0e/src/resnet.jl#L66. Its difficult to make sense of the weight names manually.

ayush1999:
It's strange that ONNX conv operator tests passed without cross-correlation. We should verify if this is indeed the correct approach. (This also may be the reason behind ONNX models not giving proper results.)
In case this is correct, and we need to flip the weights, I would like to open an issue in ONNX to verify if the test data they've provided is accurate.
cc @MikeInnes

MikeInnes:
Great that this it working. I'll merge this for now while we work on a better fix.

@ayush1999 see what happens with the ONNX tests with the flip. If we need it we should really do it at import time, either by flipping the weights then or by asking Flux to do cross-correlation.

ayush1999:
@avik-pal I've fixed the batchnorm error in ONNX. Ping me if you need the changes for other models.

PR #17, opened 2018-07-14
Adding some Untrained Models

avik-pal:
I have added the vgg nets (with and without BN). The current API update is
`VGG11()` fetches VGG11 but without pretrained weights and the model is in `train mode`.
calling `trained(VGG19)` will give the pretrained VGG19 model in the `test mode`. All models that lack pretrained weights will throw an error if called with `trained`.
@MikeInnes could you review this new changes. I will add the remaining models if these changes are ok.


src/model.jl
```diff
@@ -1,3 +1,23 @@
 abstract type ClassificationModel{Class} end
 labels(model::Type{<:ClassificationModel{Class}}) where {Class} = labels(Class)
 labels(model::ClassificationModel{Class}) where {Class} = labels(Class)
+
+function trained(which)
```
MikeInnes: Seems like you could just use dispatch with an error in the fallback case here?

src/vgg.jl
```diff
@@ -0,0 +1,156 @@
+function trained_vgg19_layers()
+  ws = weights("vgg19.bson")
+  ls = Chain(
+    Conv(ws[:conv1_1_w_0][end:-1:1,:,:,:][:,end:-1:1,:,:], ws[:conv1_1_b_0], relu, pad = (1,1), stride = (1,1), dilation = (1,1)),
+    Conv(ws[:conv1_2_w_0][end:-1:1,:,:,:][:,end:-1:1,:,:], ws[:conv1_2_b_0], relu, pad = (1,1), stride = (1,1), dilation = (1,1)),
+    x -> maxpool(x, (2,2)),
```
MikeInnes: Would be nice to use Flux's new MaxPool layer here

src/vgg.jl
```diff
@@ -0,0 +1,156 @@
...
+
+VGG11() = VGG11(load_vgg(vgg_configs["vgg11"]))
+
+Base.show(io::IO, ::VGG11) = print(io, "VGG11()")
+
+Flux.treelike(VGG11)
+
+(m::VGG11)(x) = m.layers(x)
+
+struct VGG11_BN <: ClassificationModel{ImageNet.ImageNet1k}
```
MikeInnes: Rather than having `_BN`, it'd be nicer to call `VGG19(batchnorm = true/false)`.

avik-pal:
Should I consider removing some of the tests, since travis is getting OutOfMemory Error?. I will fix the other conv error, the test follows the old API

staticfloat:
It's weird that it's running out of memory so soon; that's literally just constructing the very first model, and it's smaller than the models we've been training on in the past.  That makes me think there's something extremely memory-inefficient in how you're creating the models or something?!

avik-pal:
Yeah, it's quite strange. The test fails for the batchnorm models. I am not quite sure how to loading in a more memory efficient way as it is simply constructing an array of the layers and making a Chain out of it. And even weird the test is passing for the ResNet152 and Densenet264 which are much more memory consuming.
I am not sure if batchnorm is somehow at fault, because the densenet and resnets employ batchnorm as well.

avik-pal:
Looking at the tests I am using, it seems that the previous model stays in memory while I'm trying to generate a new one. I think we should deallocate that before building the next model.

staticfloat:
Ah, I thought it was failing on the untrained model tests.  It's the trained model tests.  Yes, that makes much more sense.  Try putting a call in to `GC.gc()` at the end of every model run?

staticfloat:
Interesting.  The OSX builder got through, but the Linux builder didn't.

staticfloat:
Okay I think what's going on here is that when we call `GC.gc()` at the end of the loop, technically we still have `model` available in scope, so DenseNet 264 can't be freed, then we immediately head off to build VGG11, and DenseNet 264 + VGG11 is too much.  Try putting a `GC.gc()` between the two loops, and also put a `whos()` in to see what else is using memory.

avik-pal:
Wasn't `whos` replaced with `varinfo`, atleast thats what my local julia07 is saying.

avik-pal:
varinfo() prints empty results yet it runs out of memory!! I can't seem to reproduce this outofmemory error locally or remotely even on another CPU machine.

staticfloat:
Let's put debugging information within the constructor; e.g. print out the sizes of the `Dense` layers as you're creating them, print out hte sizes of the `Conv()` layers as you're creating them, etc...

If that doesn't work, let's start a debug job on Travis, SSH in, and run it there.  [Supposedly we have 4GB of memory to work within](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system).

avik-pal:
Do you mean putting print statements [here](https://github.com/FluxML/Metalhead.jl/blob/145027553c3fc54091de9b685c74464a05ca92c8/src/vgg.jl#L41)?

staticfloat:
@avik-pal I couldn't figure out the memory bug; I improved memory usage, but it's still not enough to fit within 4GB.  It should fit within 7.5GB though, so I changed the Travis testing configuration to get us there.  Once we have better tools for memory management within Julia (something I'm sure we will build eventually) we can revisit this, but for now let's not spend too much time trying to track this down.

staticfloat:
If this goes green, feel free to merge whenever you see fit.

avik-pal:
Now failing for the VGG19 trained.

avik-pal:
Also we should postpone merge till the squeezenet and the googlenet have the same updated API.

avik-pal:
Now all the models have a common interface.

staticfloat:
Just as an update on this, I am currently attempting to hack together an allocation profiler so that we can see where the memory is going within Julia.  This test suite should really not be using as much memory as it is, and combining that with our GPU memory problems, I think we really just need better tools to be able to see what's going on, so I hope to (in a view days) have something hooked up with the `Profile` module that will allow us to see something similar to the flame graphs that ProfileView spits out, but showing where large objects are being allocated, and how long they survive.

src/resnet.jl
```diff
@@ -50,17 +50,38 @@ function trained_resnet50_layers()
...
   weights = Dict{Any ,Any}()
   for ele in keys(weight)
-    weights[string(ele)] = convert(Array{Float64, N} where N, weight[ele])
+    weights[string(ele)] = weight[ele]
   end
   ls = load_resnet(resnet_configs["resnet50"]...)
-  ls[1].weight.data .= flipkernel(weights["gpu_0/conv1_w_0"])
+  ls[1][1].weight.data .= flipkernel(weights["gpu_0/conv1_w_0"])
+  ls[1][2].σ² .= weights["gpu_0/res_conv1_bn_riv_0"]
+  ls[1][2].layers[2].μ .= weights["gpu_0/res_conv1_bn_rm_0"]
```
ekinakyurek: @avik-pal I didn't understand this line. If ls[1][2] isa BatchNorm layer, then how/why do you put `.layers[2]` here ?

avik-pal: That's a copy-paste error.

avik-pal:
@staticfloat any update on the status of the allocation profiler?

staticfloat:
Amazingly enough, [YES](https://github.com/JuliaLang/julia/pull/31534)!  I don't have time to apply the memory profiler to this branch immediately, but in the next week or two I hope to.  It may shed some light on what's going on here.  If someone else want to do this, I suggest telling the memory profiler to only pay attention to `big` and `std` allocations, as anything small enough to fit within the pool is likely not useful to our analysis here.

PR #18, opened 2018-07-18
Fix VGG19 preprocessing

staticfloat:
This works a lot better for me, and most importantly, classifies as `malamute` for Alan's `phillip.jpg`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.6
 Flux
 Images
+ImageFiltering
 BSON
--- a/src/Metalhead.jl
+++ b/src/Metalhead.jl
@@ -1,7 +1,7 @@
 __precompile__()
 module Metalhead
 
-using Flux, Images, BSON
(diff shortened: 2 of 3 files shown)


staticfloat:
@MikeInnes the only weird part of this is that I have to enforce `Float32` for the input data, since the model weights are `Float32`.  If I don't do that, I get an error that a `conv` method isn't defined for the given types.  My guess is that somewhere in `NNlib` there is a `conv()` method that doesn't allow kernel and input types to be different.

It's also possible that this has been fixed on NNlib master, I'm on cf9e186 since I'm on 0.6.

staticfloat:
> I guess we should add `ImageFiltering` to require?

:+1:

staticfloat:
I can't restart the travis build, but it looks like an unrelated error.  Can we rerun then merge if this works?  We should really update the title picture.

MikeInnes:
Thanks @staticfloat!

PR #19, opened 2018-08-26
Julia 1.0 updates

avik-pal:
I have tested the working of the models and the dataset imports.
@ayush1999 could you see why the `squeezenet` is not working? It is giving a segfault.
Also MikeInnes/BSON.jl#10 needs to be merged for this to work
cc @Keno


ayush1999:
I'll look into it. Also, @avik-pal  are you sure these changes work at this moment? I'm getting a few errors while importing the package.

avik-pal:
You need to checkout the working version of BSON which for which I provided the link. You might encounter some warnings while importing but that is due to Flux overwriting the Broadcast in certain places

MikeInnes:
BSON is now tagged and the warning from Flux is fixed. We can merge this without squeezenet if we have to.

ayush1999:
@MikeInnes I think you can merge this for now. I'll investigate the reason behind squeezenet failure and open a separate PR.

MikeInnes:
The build still seems to be failing on travis and locally for me, despite squeezenet being commented out.

avik-pal:
Even VGG is giving a `segfault`

MikeInnes:
Ok, VGG is pretty essential for this. Can you narrow down what op is causing that?

ayush1999:
Just realized, all the ONNX models are working on julia0.6, but give a segfault on Julia0.7. @MikeInnes Any idea what could be causing this to happen?

MikeInnes:
If I had to guess, something has changed in NNlib and one of the kernels that does `@inbounds` is broken. But you'll have to bisect the model to narrow it down to a single function call to verify that.

ayush1999:
Looks like I found the reason behind these crashes. I've opened an issue in the Flux repo (https://github.com/FluxML/Flux.jl/issues/380)

avik-pal:
Should we just convert the weights to Float64 (for vgg and squeezenet) and update Metalhead to 1.0 for now, till we can figure out the solution for Float32?

PR #25, opened 2018-09-17
Fix deprecated functions (method_exists, Flux.argmax)

wookay:
Fixed some deprecated functions
- Warning: `method_exists(f, t)` is deprecated, use `hasmethod(f, t)` instead.
- Warning: `argmax(...)` is deprecated, use `onecold(...)` instead.


MikeInnes:
:+1: Thanks!

Issue #28, opened 2018-10-03
MethodError: no method matching conv(::Array{Float32,4}, ::Array{Float64,4}; ...)


c-p-murphy:
When I was playing around with `Metalhead` I came across this error and I'm not sure how to fix it. I think it has to do with one of the arguments being an `Array{Float32,4}` and the other an `Array{Float64,4}` The same error happens with `ResNet` and `DenseNet`, but `VGG19` and `SqueezeNet` work fine.

```julia
julia> x = rand(Float32, 224, 224, 3, 1)
224×224×3×1 Array{Float32,4}:
...

julia> gn = GoogleNet()
GoogleNet()

julia> gn(x)
ERROR: MethodError: no method matching conv(::Array{Float32,4}, ::Array{Float64,4}; stride=(2, 2), pad=(3, 3), dilation=(1, 1))
Closest candidates are:
  conv(::AbstractArray, ::TrackedArray; kw...) at /Users/christophermurphy/.julia/packages/Flux/UHjNa/src/tracker/array.jl:305
  conv(::TrackedArray, ::AbstractArray; kw...) at /Users/christophermurphy/.julia/packages/Flux/UHjNa/src/tracker/array.jl:306
  conv(::A<:AbstractArray, ::A<:AbstractArray; pad, stride, dilation) where A<:AbstractArray at /Users/christophermurphy/.julia/packages/NNlib/0EAe7/src/conv.jl:28
Stacktrace:
 [1] #_forward#335(::Base.Iterators.Pairs{Symbol,Tuple{Int64,Int64},Tuple{Symbol,Symbol,Symbol},NamedTuple{(:stride, :pad, :dilation),Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64},Tuple{Int64,Int64}}}}, ::Function, ::typeof(NNlib.conv), ::Array{Float32,4}, ::TrackedArray{…,Array{Float64,4}}) at /Users/christophermurphy/.julia/packages/Flux/UHjNa/src/tracker/array.jl:308
 [2] (::getfield(Flux.Tracker, Symbol("#kw##_forward")))(::NamedTuple{(:stride, :pad, :dilation),Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64},Tuple{Int64,Int64}}}, ::typeof(Flux.Tracker._forward), ::typeof(NNlib.conv), ::Array{Float32,4}, ::TrackedArray{…,Array{Float64,4}}) at ./none:0
[...]

PR #29, opened 2018-10-05
Add option for Type in preprocess

avik-pal:
Now that we have some models having `Float32` and `Float64` weights we should allow the datatype of the array returned by preprocess function to be changed from outside. This problem was raised in #28.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -47,7 +47,7 @@ function center_crop(im, len)
   return im[div(end,2)-l2:div(end,2)+l2-adjust,div(end,2)-l2:div(end,2)+l2-adjust]
 end
 
-function preprocess(im::AbstractMatrix{<:AbstractRGB})
+function preprocess(T::DataType, im::AbstractMatrix{<:AbstractRGB})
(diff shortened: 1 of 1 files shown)


MikeInnes:
This is a good change, but in general we should just make the conv api a bit more flexible. I think there are prs open for that.

avik-pal:
In that case, I am closing this PR :)

MikeInnes:
It would be fine to do this too, they are not exclusive options :) see what @staticfloat thinks.

staticfloat:
At this point, I think it's cleaner for us to make `Conv()` more flexible, then completely remove the `Float32.()` in the `return` statement.  Let's keep data types native as long as we can, only widening once an operator requires it of us.

ekinakyurek:
I don't understand the point. In current state, Conv operation wants parameters and the inputs to be  in same precision. It is logical because you want your gradients to be in the same precision with your parameters and you also want that Conv model respects you're input precision. So, it cannot be done at the same time. 

After that discussion, why you don't simply set both model parameters and inputs in Metalhead  to Float64 or to Float32 ?

PR #34, opened 2019-01-31
allow preprocessing grayscale images

KristofferC:


MikeInnes:
This still seems like a good idea to me but it'd just need a look over from someone like @staticfloat.

Should it be `AbstractMatrix{<:Real}`?

MikeInnes:
Seems like it can't hurt. Thanks @KristofferC!

PR #40, opened 2019-03-14
Added PASCAL VOC dataset for object detection

shreyas-kowshik:


darsnack:
Closing this as Metalhead.jl no longer ships datasets, only models. MLDatasets.jl will be a good home for this work.

PR #48, opened 2019-10-01
use Pkg3

dhairyagandhi96:
Ref also FluxML/model-zoo#173


dhairyagandhi96:
Woo.

Would like to merge this if there are no objections.

PR #70, opened 2020-08-28
Refactor repo with more complete models and documentation

darsnack:
~~**WIP: DO NOT MERGE!**~~

This is a major refactor to Metalhead.jl with the following changes:
- Datasets are removed
- Classification/prediction logic is removed
- Updated VGG, GoogLeNet, ResNet to code from FluxModels.jl
- Updated DenseNet, SqueezeNet (post-FluxModels.jl)
- Added AlexNet, Inception v3

Overall, the repository will now operate as the de-facto source of vision models for the Flux ecosystem (similar to `torchvision.models`). The codebase has been slimmed down so that we can provide users with a lightweight dependency to get standard pre-trained models. The models are also updated to use the latest layers in Flux like `SkipConnection`, `Parallel`, and adaptive pooling.

Most importantly, the model is code is now more flexible. For example, the ResNet code is generic enough that users can extend it to create variants of ResNet that are not in the original paper. This is useful for research where varying model architecture parameters is important (e.g. double descent). We also provide **all** the ResNet variants from the paper, not just ResNet-50 like most other ecosystems. The same can be said for other models like VGG, etc.

Future work:
- [x] Add pre-trained weights
    - [x] ~~alexnet~~ (future PR)
    - [x] VGG
        - [x] ~~vgg11~~ (future PR)
        - [x] ~~vgg11bn~~ (future PR)
        - [x] ~~vgg13~~ (future PR)
        - [x] ~~vgg13bn~~ (future PR)
        - [x] ~~vgg16~~ (future PR)
        - [x] ~~vgg16bn~~ (future PR)
        - [x] vgg19
        - [x] ~~vgg19bn~~ (future PR)
    - [x] ResNet
        - [x] ~~resnet18~~ (future PR)
        - [x] ~~resnet34~~ (future PR)
        - [x] resnet50
        - [x] ~~resnet101~~ (future PR)
        - [x] ~~resnet152~~ (future PR)
    - [x] googlenet
    - [x] inceptionnet
    - [x] squeezenet
    - [x] DenseNet
        - [x] densenet_121
        - [x] ~~densenet_161~~ (future PR)
        - [x] ~~densenet_169~~ (future PR)
        - [x] ~~densenet_201~~ (future PR)
- [x] Add documentation
- [x] Tests

This PR will supersede #69. cc @DhairyaLGandhi


darsnack:
Currently tests are failing without https://github.com/FluxML/Flux.jl/pull/1305

dmolina:
Nice, great work!. However, because there is missing functionallity from previous version, I think it could be better to update it in another branch,  and when the missing functionality (specially the pre-trained weights and missing models) merge it with master. It is a topic of style, but I way it to avoid problems to people that could be using #master version. Don't think I don't believe that MetalHead should be changed, I'm convinced that it should be clearly improved.But we must avoid temporarily generating problems to users.

darsnack:
I guess I can't mark a PR as a draft after it's already been created? But I added a disclaimer at the top of the PR. I agree, we shouldn't merge this into `master` until it restores all the lost functionality.

DhairyaLGandhi:
(There should be a little text field under the reviewers area marked "convert to draft")

DhairyaLGandhi:
Why is this removing densenet and squeezenet?

darsnack:
It isn't. I just haven't updated the PR to have them which is why I marked it as WIP.

src/inception.jl
```diff
@@ -0,0 +1,150 @@
+function conv_block(kernelsize::Tuple{Int64,Int64}, inplanes::Int64, outplanes::Int64; stride::Int64=1, pad::Union{Int64,Tuple{Int64,Int64}}=0)
+    conv_layer = []
+    push!(conv_layer, Conv(kernelsize, inplanes => outplanes, stride=stride, pad=pad))
+    push!(conv_layer, BatchNorm(outplanes, relu))
+    return conv_layer
+end
```
darsnack, line 6: It seems like this is a more general version of `conv_block` in "googlenet.jl". Can we add a "utilities.jl" file to collect utility functions like these that can be used across models? I think also renaming the function to `conv_bn` will be clearer.

Also, instead of creating an array, maybe we can just return a tuple?

darsnack, line 6: It seems like this is a more general version of `conv_block` in "googlenet.jl". Can we add a "utilities.jl" file to collect utility functions like these that can be used across models? I think also renaming the function to `conv_bn` will be clearer.

Also, instead of creating an array, maybe we can just return a tuple?

src/resnet.jl
```diff
@@ -1,97 +1,100 @@
...
+        Conv((3, 3), outplanes[1] => outplanes[2], stride = 1, pad = 1),
+        BatchNorm(outplanes[2], relu),
+        Conv((1, 1), outplanes[2] => outplanes[3], stride = 1),
+        BatchNorm(outplanes[3], relu)) :
+  Chain(Conv((1, 1), inplanes => outplanes[1], stride = 1),
+        BatchNorm(outplanes[1], relu),
+        Conv((3, 3), outplanes[1] => outplanes[2], stride = 1, pad = 1),
+        BatchNorm(outplanes[2], relu),
+        Conv((1, 1), outplanes[2] => outplanes[3], stride = 1),
+        BatchNorm(outplanes[3], relu))
```
darsnack, line 23: Could these be updated to use the new `conv_bn` utility function?

src/resnet.jl
```diff
@@ -1,97 +1,100 @@
...
+
+function projection(inplanes, outplanes, downsample = false)
+  shortcut = downsample ? 
+    Chain(Conv((1, 1), inplanes => outplanes, stride = 2),
+          BatchNorm((outplanes), relu)) :
+    Chain(Conv((1, 1), inplanes => outplanes, stride = 1),
+          BatchNorm((outplanes), relu))
```
darsnack, line 30: Same comment here

src/resnet.jl
```diff
@@ -1,97 +1,100 @@
...
-  for i in 1:length(layers)
-    push!(layer_arr, Bottleneck(initial_filters, true, i==1))
-    for j in 2:layers[i]
-      push!(layer_arr, Bottleneck(initial_filters))
+function resnet(block, shortcut_config, channel_config, block_config)
+  inplanes = 64
+  baseplanes = 64
+  layers = []
+  push!(layers, Conv((7, 7), 3=>inplanes, stride=(2, 2), pad=(3, 3)))
+  push!(layers, BatchNorm(inplanes, relu))
```
darsnack, line 53: Same comment

src/vgg.jl
```diff
@@ -0,0 +1,90 @@
...
+function vgg_block(ifilters, ofilters, depth, batchnorm)
+  k = (3,3)
+  p = (1,1)
+  layers = []
+  for l in 1:depth
+    if batchnorm
+      w = convfilter(k, ifilters=>ofilters)
+      b = Zeros()
+      push!(layers, Conv(weight=w, bias=b, pad=p))
+      push!(layers, BatchNorm(ofilters, relu))
```
darsnack, line 16: Same comment

darsnack (commented):
@rishabhvarshney14 looks good. I added a couple of comments to use utility functions instead of repeating code.

Are you interested in working on the other remaining models (SqueezeNet and DenseNet)?

dmolina:
A comment about the commit about inception, thank you for that, sorry but I do not know how to comment the commit, I think the test has a small error, because in the test the model is googlenet, not inception.

darsnack:
Good catch! @rishabhvarshney14 can you address the test error too?

rishabhvarshney14:
@dmolina  Thanks for noticing I have fixed that.
@darsnack Sure I can work on other models too and also make the required changes Thank You.

darsnack:
Awesome, thank you!

src/vgg.jl
rishabhvarshney14: Just a quick question, here Conv layer takes directly weight and bias and b = Flux.Zeros(). According to [Flux.jl's docs](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Conv) "Setting bias to Flux.Zeros() will switch bias off for the layer."

Should I modify 'conv_bn' function for this or just use the same conv_bn function here?

darsnack: Maybe add a `usebias = true` keyword argument for `conv_bn`? Then we can use it here with `usebias = false` and ignore everywhere else.

src/densenet.jl
```diff
@@ -1,92 +1,57 @@
...
+function bottleneck(inplanes, growth_rate)
+    inner_channels = 4 * growth_rate
+    m = Chain(conv_bn((1, 1), inplanes, inner_channels; usebias=false, rev=true)...,
+              conv_bn((3, 3), inner_channels, growth_rate; pad=1, usebias=false, rev=true)...)
 
-(b::Bottleneck)(x) = cat(b.layer(x), x, dims = 3)
+    layer = x -> begin
+        y = m(x)
+        return cat(x, y; dims=3)
+    end
```
darsnack, line 11: I would rather￼ return a `SkipConnection` here

src/densenet.jl
```diff
@@ -1,92 +1,57 @@
-struct Bottleneck
-    layer
-end
-
-@functor Bottleneck
+using Flux
```
darsnack, line 1: Shouldn't need this line

src/densenet.jl
```diff
@@ -1,92 +1,57 @@
...
-    local layers = []
+function make_dense_block(inplanes, growth_rate, nblock)
+    layers = []
     for i in 1:nblock
-        push!(layers, block(in_planes, growth_rate))
-        in_planes += growth_rate
+        push!(layers, bottleneck(inplanes, growth_rate))
+        inplanes += growth_rate
     end
     Chain(layers...)
```
darsnack, line 24: By convention, we haven't been returning intermediate `Chain`s. Here you can just return `layers`.

src/densenet.jl
```diff
@@ -1,92 +1,57 @@
...
+    layer = x -> begin
+        y = m(x)
+        return cat(x, y; dims=3)
+    end
+end
 
-Transition(chs::Pair{<:Int,<:Int}) = Chain(BatchNorm(chs[1], relu),
-                                            Conv((1, 1), chs),
-                                            MeanPool((2, 2)))
+transition(inplanes, outplanes) = Chain(conv_bn((1, 1), inplanes, outplanes; usebias=false, rev=true)...,
```
darsnack, line 14: Same thing here. Return a vector instead of the `Chain`.

src/vgg.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+# Build convolutionnal layers
+#  config: :A (vgg11) :B (vgg13) :D (vgg16) :E (vgg19)
+#  inchannels: number of channels in input image (3 for RGB)
+function convolutional_layers(config, batchnorm, inchannels)
+  layers = []
+  ifilters = inchannels
+  for c in config
+    push!(layers, vgg_block(ifilters, c..., batchnorm)...)
```
darsnack, line 29: I guess we can `append!` here to avoid splatting (sorry minor change).

darsnack (commented):
Thank you! It looks great. After these changes, I'll spin some cycles on my GPU to add pre-trained weights.

darsnack (commented):
Looks good, thanks. I'll add the pre-trained weights, and then we can merge this PR!

cyborg1995:
@darsnack, referencing [this comment](https://github.com/FluxML/ML-Coordination-Tracker/issues/2#issuecomment-744094102), do you mean to train the models from scratch on Imagenet dataset myself or writing the code for training?

Since it is such a massive dataset, It would be difficult for me to train it myself but I can surely write the code which can be used by you/another contributor with the appropriate resources to train the model.

darsnack:
Yeah even the code would help. Can you run for a few epochs? If that's feasible, then just doing that and posting a gist of the script that trains the models will allow me to just run that script on GPU machine. Then we'd have the trained weights in short order.

cyborg1995:
@darsnack, I have created the [gist](https://gist.github.com/cyborg1995/168826876af16b18640ec4f7b95e827b) for training alexnet using the architecture [here](https://github.com/darsnack/Metalhead.jl/blob/darsnack/vision-refactor/src/alexnet.jl) but haven't tried running it though.
Please try to train it on your end and let me know if it works (or not!)
Currently, checkpointing the model weights is not implemented.
The weights are saved only after the model has completed training.

darsnack:
Thanks @cyborg1995, I'll give it a go tonight and report back.

rishabhvarshney14:
@darsnack @cyborg Sorry if I misunderstand this but in [Transformers.jl](https://github.com/chengchingwen/Transformers.jl) it uses pretrained weights of Hugging Face models to provide pretrained models in Julia/Flux. It might be better if we use pretrained weights of other models and use it here instead of training it from the start.

darsnack:
@rishabhvarshney14 Unfortunately, that isn't a simple translation since a Flux model and a PyTorch model aren't the same structure.

I think it's an easier "once and done" to just train from scratch. It's just a question of someone with GPUs (e.g. me) to have the bandwidth to do it.

darsnack:
Actually I take that back. There would be a lot of parsing work involved to map the pre-trained weights from PyTorch to the corresponding layers in the Flux models. If someone wants to attempt to tackle that, then that would be great. Otherwise, I'll have to find some time to train these models later this weekend.

cyborg1995:
@darsnack, isn't it possible to convert the pre-trained ONNX models to Flux models using [ONNX.jl](https://github.com/FluxML/ONNX.jl)

darsnack:
Not quite. It's easy to read the weights in with ONNX.jl, but since ONNX serializes to [more primitive operations](https://github.com/onnx/onnx/blob/master/docs/Operators.md), you need some translation code to figure out which higher level Flux layer contains the convolution etc. that you read from ONNX.

cyborg1995:
I think doing it from ONNX would be easier. I'll try to figure it out.

darsnack:
@cyborg1995 You might look at [ONNXmutable.jl](https://github.com/DrChainsaw/ONNXmutable.jl) which is more up to date/maintained. If you do figure out some translation code that works, then do you mind sharing it in [this issue](https://github.com/FluxML/ML-Coordination-Tracker/issues/10)? Having a good translation flow from ONNX to Flux is one of our goals.

darsnack:
Another actionable item here for anyone looking to contribute: we will probably add the BSON weight files using [Pkg.jl artifacts](https://julialang.github.io/Pkg.jl/v1/artifacts/). If someone can set the repo up to do that, then please PR my fork with your code.

Additionally, once we have artifacts in place, we need to update the model constructors with a `pretrain = false` keyword argument that will optionally call `BSON.@load` on the respective weight file. Again, feel free to PR my fork if you can get this up and running.

darsnack:
An update on this PR. I have updated the code to use the latest features from Flux like the `Parallel` layer. I have also added a `training` folder that contains scripts to obtain the pre-trained weights. There are instructions on how to use the scripts under `training/README.md`. The single GPU training works, but the distributed training is giving me issues. I don't think that's a problem, since we can easily parallelize by training multiple models at once on different devices.

As expected, training on ImageNet is time consuming, and I am hoping that we can make this a collective effort. If you have GPUs to spare, please clone my fork and train a model. If you need to obtain ImageNet, DM me on Slack or Zulip so we can sort it out. Everything should work out of the box.

**Here is the process for training a model:**
1. Check the top post of this PR to see which models need training
2. Claim your model by commenting so that we don't train the same model twice
3. Submit a PR to my fork with the committed BSON

According to torchvision, here are the accuracies we need to hit on ImageNet's validation data:
Screen Shot 2021-01-24 at 12 06 54 PM

cc @ToucheSir @lorenzoh

DhairyaLGandhi:
Training things really does need an api and some overloads to flux seem extraneous. I'm assuming a lot of this needs major cleaning up still?

DhairyaLGandhi:
All the trained weights need to become artifacts, so best not to have them in a fork, we will need to update them anyway. I can help with that.

Also we can train multiple models at once

DhairyaLGandhi:
Happy to coordinate on that

darsnack:
Yeah I didn't know where you wanted to host the artifacts. Let me know so we can figure out the best way for folks to send us the BSON.

The training directory isn't an API, nor is it meant to be used by users. It's completely internal for us to generate the weights (note that it's not under `src`). We can even delete it once the weights are generated, though I think there is value in keeping it around.

darsnack:
To clarify, the training folder doesn't need cleaning up. It should work as it currently exists. Yes, I use a custom `train!` but that's because I need a LR scheduler to step every epoch. Nothing else overloads Flux.

lorenzoh:
Since the models should reflect the best practices and provide parity with other deep learning frameworks, I would suggest to replace Flux.jl's default initializer for `Conv` (`glorot_uniform`) with `Flux.kaiming_normal` for these models or at least the ResNets as is common practice. Can confirm from anecdotal experience that it gives slightly better convergence.
See [the PyTorch implementation](https://pytorch.org/vision/stable/_modules/torchvision/models/resnet.html). [fastai also uses `kaiming_normal`](https://github.com/fastai/fastai/blob/66a03da8a11cd85188f4c6f063b98c4a209492e8/fastai/vision/models/xresnet.py#L18).

Maybe this should become the default for Flux.jl at some point but maybe that is too large of a breaking change.

src/utilities.jl
```diff
@@ -0,0 +1,20 @@
+function conv_bn(kernelsize::Tuple{Int64,Int64}, inplanes::Int64, outplanes::Int64;
+                 stride::Int64=1, pad::Union{Int64,Tuple{Int64,Int64}}=0,
+                 usebias::Bool=true, rev::Bool=false)
+  conv_layer = []
+  if usebias
+    push!(conv_layer, Conv(kernelsize, inplanes => outplanes, stride=stride, pad=pad))
```
lorenzoh, line 6:
```suggestion
    push!(conv_layer, Conv(kernelsize, inplanes => outplanes, stride=stride, pad=pad, init=Flux.kaiming_normal))
```

src/utilities.jl
```diff
@@ -0,0 +1,20 @@
...
+  else
+    push!(conv_layer, Conv(kernelsize, inplanes => outplanes, stride=stride, pad=pad, bias=Flux.Zeros()))
```
lorenzoh, line 8:
```suggestion
    push!(conv_layer, Conv(kernelsize, inplanes => outplanes, stride=stride, pad=pad, bias=Flux.Zeros(), init = Flux.kaiming_normal))
```

lorenzoh (changes requested):

darsnack:
Moving this out of a draft, cause we're nearing the end (at the very least some eyes for code review would be good).

A couple of remaining items:
- documentation
- tests for loading pre-trained weights

Also waiting on the Flux v0.12 release since some of these models use `Parallel`.

darsnack:
@CarloLucibello @DhairyaLGandhi I think this is ready for review. Tests, docstrings, and a quick tutorial are added. We can't merge until Flux v12 is released though. Only minor changes from here on (e.g. addressing @lorenzoh suggestion above).

Project.toml
```diff
@@ -1,28 +1,21 @@
...
 
 [compat]
-BSON = "0.2"
-ColorTypes = "0.8, 0.9"
-Flux = "0.10, 0.11"
-ImageFiltering = "0.6"
-Images = "0"
-Requires = "0.5, 1.0"
+BSON = "0.3.2"
 julia = "1.4"
```
CarloLucibello, line 12: should add Flux v0.12 as lower bound?

docs/tutorials/quickstart.md
```diff
@@ -0,0 +1,30 @@
...
+using Flux: onehotbatch
+using Statistics: mean
+
+batchsize = 8
+data = [(rand(Float32, 224, 224, 3, batchsize), onehotbatch(rand(1:1000), 1:1000))
+        for _ in 1:3]
+opt = ADAM()
+ps = Flux.params(model)
+loss(x, y, m) = Flux.Losses.logitcrossentropy(m(x), y)
+cb = () -> @show mean(loss(x, y, model) for (x, y) in data)
```
CarloLucibello, line 28: the callback is called after each gradient step, so evaluating on the whole dataset it's not good even if this is an example with dummy data. Use manual loop instead?

CarloLucibello:
Besides those two comments looks good to me. We should hurry up with the v0.12 release

Project.toml
darsnack: Yes will do that once the release is tagged (had some problems withe the `-DEV`...I was probably doing something wrong).

docs/tutorials/quickstart.md
darsnack: Yeah ideally, we would not do a training example but evaluate the pre-trained model on a sample from ImageNette in MLDatasets.jl. But to include that in the docs might mean we want to spend some time fusing FastAI.jl datasets and MLDatasets.jl better (ImageNette comes from FastAI.jl).

darsnack: I plan on doing a separate training loop example with a Jupyter notebook. But that requires https://github.com/MichaelHatherly/Publish.jl/issues/9. Doesn't seem like too hard a fix, but I just haven't had the time to look into it yet. The reason for preferring Jupyter notebooks is that training these models requires a GPU, so ideally, we would save the notebook (storing the output), and then we don't have to re-run the training to build the docs.

Project.toml
```diff
@@ -1,28 +1,22 @@
 name = "Metalhead"
 uuid = "dbeba491-748d-5e0e-a39e-b530a07fa0cc"
-version = "0.5.1"
+version = "0.5.3"
```
CarloLucibello, line 3:
```suggestion
version = "0.6.0"
```

docs/make.jl
```diff
@@ -0,0 +1,15 @@
...
+Publish.Themes.default() = artifact"flux-theme"
+
+p = Publish.Project(Metalhead)
+
+# needed to prevent error when overwriting
+rm("dev", recursive = true, force = true)
+rm(p.env["version"], recursive = true, force = true)
+
+# build documentation
+deploy(Metalhead; root = "/Metalhead.jl", force = true, label = "dev")
```
CarloLucibello, line 15:
```suggestion
deploy(Metalhead; root = "/Metalhead.jl", force = true, label = "dev")

```

docs/serve.jl
```diff
@@ -0,0 +1,12 @@
...
+
+using Metalhead
+
+# override default theme
+Publish.Themes.default() = artifact"flux-theme"
+
+p = Publish.Project(Metalhead)
+
+# serve documentation
+serve(Metalhead)
```
CarloLucibello, line 12:
```suggestion
serve(Metalhead)

```

src/alexnet.jl
```diff
@@ -0,0 +1,30 @@
...
+                 Dropout(0.5),
+                 Dense(256 * 6 * 6, 4096, relu),
+                 Dropout(0.5),
+                 Dense(4096, 4096, relu),
+                 Dense(4096, 1000))
+
+  pretrain && pretrain_error("alexnet")  
+
+  return layers
+end
```
CarloLucibello, line 30:
```suggestion
end

```

CarloLucibello (approved):

CarloLucibello:
Looks great! just a rebase away from merging

DhairyaLGandhi:
Let's split this out into multiple prs. I would love to give this a proper go before we merge.

DhairyaLGandhi:
Are we dropping all the planned pretrained models here?

darsnack:
How would you want to split it up?

> Are we dropping all the planned pretrained models here?

Not all. All the models that were available as pre-trained before this PR will be available as pre-trained after this PR. Only the new models are not available as pre-trained within this PR (the ones crossed out on the top). Those will be added individually as the weights are ready.

DhairyaLGandhi:
The models and the dataset handling (maybe docs can be with the models)


We should also have the high level names as earlier. For example `ResNet` instead of `resnet`, with an arg to control the sizes.

darsnack:
There is no training or dataset handling. There are only models.

I can change the names, but the reason I went lowercase is because all the models are "functions" in that they don't return a struct called `ResNet`. Conventionally, camel case naming would be used for a struct constructor.

DhairyaLGandhi:
So I think it's still desired to have a wrapper struct for `show` ing properly, dispatch (like the kinds needed in AlphaZero) and having a compact representation that is nicely serializable to boot.

tt also allows us to take more liberties since some plumbing functions become more "internal", but thats not something we can do in general.

Another thing is variables naming. I wouldn't want users to be confused with all lowercase variables and functions and models.

darsnack:
Sure might actually be nicer with a single wrapper to collect network parameters (either in type or fields) like AlphaZero.jl

DhairyaLGandhi:
Single wrapper?

darsnack:
I mean like a `ResNet` for all the ResNets, `VGG` for all the VGGs, etc.

darsnack:
Looks like we are getting OOM errors on the Github CI (previously, Metalhead.jl used Travis). Any ideas how to address this (tests pass locally...)?

darsnack:
Okay aside from the OOM issues on CI, I think this is RTM.

Artifacts.toml
```diff
@@ -0,0 +1,39 @@
+[densenet121]
+git-tree-sha1 = "ffc7f7ed1e7f67baca4b76f6c100e0d5042ff063"
+lazy = true
+
+    [[densenet121.download]]
+    sha256 = "3fd10f0be70cf072fa7f1358f1fbbe01138440dbcaec1b7c8e007084382c1557"
+    url = "https://github.com/darsnack/MetalheadWeights/releases/download/v0.1.1/densenet121-0.1.1.tar.gz"
```
ToucheSir, line 7: Will this repo be moved into FluxML or incorporated into Metalhead somehow?

ToucheSir (commented):
Generally LGTM on a quick skim, do you foresee any follow-up work after this PR (e.g. adding more pre-trained weights)?

DhairyaLGandhi:
Happy to review it. Do we have an updated table with the validation accuracies with imagenet with the models?

Artifacts.toml
darsnack: Done.

darsnack:
Given the latest ML call regarding the validation accuracies, I think we should go ahead with this PR as is. Once the multi-GPU demo is publicly available, I will re-train _all_ the models and update the weights. If we merge now, then at least people can use the architectures easily by adding `master`.

ablaom:
@darsnack I wonder if a final suggestion is not too late: Add a RNG to the signature of the new generic `vgg` and `resnet` constructors, for use in layer initialisation (and ultimately Dropout, assuming https://github.com/FluxML/Flux.jl/issues/1617 is accepted).

I would like to integrate these very useful constructors into MLJFlux: https://github.com/FluxML/MLJFlux.jl/issues/162 which now allow "builders" to access a user-specified RNG for our models.

darsnack:
No, it's not too late. My first inclination is to define a keyword arg like ￼`init` which is how Flux handles custom initialization including user-specified RNGs. This would allow more than just swapping RNGs — a user could switch from `kaiming_normal` to `glorot_uniform`, for examples.

This would not apply to `Dropout`, because there the RNG is affecting runtime and not initialization. To me, that's a conceptual discrepancy, and I would want to keep initialization vs. runtime separate.

Especially since the dropout RNG PR hasn't landed, what I suggest is just sticking to just `init` for now (to maintain Flux's semantics). Then for `Dropout`, we have:
1. Introduce something like `flux_rng() = use_cuda[] ? CUDA.default_rng() : Random.GLOBAL_RNG`. Pair this with a scoping block like `withrng(rng) do ... end` that overrides `flux_rng` within the do-block.
2. Store runtime RNGs within the layers (so they can be specified on construction). Have something like `swaprng(rng)` that works similar to `gpu` (i.e. `m = m |> swaprng(rng)`) that replaces all the stored RNGs with a user-specified one.

Option (2) seems like something MLJFlux could utilize in the builder. So the `vgg` builder in MLJFlux could have and `rng` keyword and calls `Metalhead.vgg` with `init` specified using `rng`. It would further call `swaprng(rng)` on the returned model to also swap the runtime RNGs. This would let Flux keep its current semantics while allowing MLJ to have the more user-friendly high level for specifying randomness.

Sorry for the long explanation, but would that be an acceptable solution?

CarloLucibello (approved):

darsnack:
FYI, I can't merge this PR (unless this repo uses `bors`). Who has merge rights on this repo?

CarloLucibello:
@DhairyaLGandhi can we have write access to this repo?

@ablaom @darsnack I would postpone the rng thing to a follow up PR

darsnack:
> I would postpone the rng thing to a follow up PR

@ablaom I think this is a good idea given how long this PR has been around. I'll open a new PR on top of this so we can track the RNG stuff.

ablaom:
@darsnack I agree that postponing the RNG PR is reasonable. Please copy me into the new PR or issue when it's ready. 

> Especially since the dropout RNG PR hasn't landed, what I suggest is just sticking to just init for now (to maintain Flux's semantics). 

Agreed.

> Then for Dropout, we have:

> 1. Introduce something like flux_rng() = use_cuda[] ? CUDA.default_rng() : Random.GLOBAL_RNG. Pair this with a scoping block like withrng(rng) do ... end that overrides flux_rng within the do-block.

> 2. Store runtime RNGs within the layers (so they can be specified on construction). Have something like swaprng(rng) that works similar to gpu (i.e. m = m |> swaprng(rng)) that replaces all the stored RNGs with a user-specified one.

Yeah, probably 2 is more MLJFlux friendly.

lorenzoh:
*push* on this PR. @DhairyaLGandhi can you merge? I think other people active here lack the permissions.

DhairyaLGandhi:
On it, will review it in its current state and move

src/densenet.jl
```diff
@@ -1,92 +1,188 @@
...
-                                          Conv((3, 3), 4growth_rate => growth_rate, pad = (1, 1))))
+Create a DenseNet transition sequence
+([reference](https://arxiv.org/abs/1608.06993)).
 
-(b::Bottleneck)(x) = cat(b.layer(x), x, dims = 3)
+# Arguments
+- `inplanes`: number of input feature maps
+- `outplanes`: number of output feature maps
+"""
+transition(inplanes, outplanes) = (conv_bn((1, 1), inplanes, outplanes; usebias=false, rev=true)...,
```
DhairyaLGandhi, line 30:
```suggestion
transition(inplanes, outplanes; bias = true, rev = true) = (conv_bn((1, 1), inplanes, outplanes; bias, rev)...,
```

src/densenet.jl
```diff
@@ -1,92 +1,188 @@
...
-
-    Chain(layers..., MeanPool((7, 7)),
-        x->reshape(x, :, size(x, 4)),
-        Dense(num_planes, num_classes), softmax)
+# Arguments
+- `inplanes`: number of input feature maps to the full sequence
+- `growth_rate`: the rate at which output feature maps grow across blocks
+- `nblock`: the number of blocks
+"""
+function dense_block(inplanes, growth_rate, nblock)
```
DhairyaLGandhi, line 44: To me its not obvious how `growth_rate` should be expressed - I'd almost expect it to work like gain and multiply. Maybe just accept a vector of values and drop `nblock`?

DhairyaLGandhi: This is of course something that would be needed in more places than just here.

src/densenet.jl
```diff
@@ -1,92 +1,188 @@
...
+"""
+    DenseNet121(; pretrain=false)
+
+Create a DenseNet-121 model
+([reference](https://arxiv.org/abs/1608.06993)).
+Set `pretrain=true` to load the model with pre-trained weights for ImageNet.
+
+See also [`Metalhead.DenseNet`](#).
+"""
+function DenseNet121(; pretrain=false)
```
DhairyaLGandhi, line 129: The nicer API would be to take an argument to specify the size of model one wants. It would be more dynamic as well about what models we produce and should help consolidate some of the code.

src/googlenet.jl
```diff
@@ -1,82 +1,88 @@
...
+# Arguments
+- `inplanes`: the number of input feature maps
+- `out_1x1`: the number of output feature maps for the 1x1 convolution (branch 1)
+- `red_3x3`: the number of output feature maps for the 3x3 reduction convolution (branch 2)
+- `out_3x3`: the number of output feature maps for the 3x3 convolution (branch 2)
+- `red_5x5`: the number of output feature maps for the 5x5 reduction convolution (branch 3)
+- `out_5x5`: the number of output feature maps for the 5x5 convolution (branch 3)
+- `pool_proj`: the number of output feature maps for the pooling projection (branch 4)
+"""
+function inceptionblock(inplanes, out_1x1, red_3x3, out_3x3, red_5x5, out_5x5, pool_proj)
```
DhairyaLGandhi, line 16: Could definitely use a nicer API here. I am not sure if we should break this down into multiple "specs" for a conv layer, but seems like we want to have access to the filter sizes, as well as manage how the reduction blocks are sized. Currently, its got too many arguments (which would also be the case if these were made kwargs, which I don't think is a good idea.)

we should also remove the `Chain` in the first block.

DhairyaLGandhi (commented):
Starting to do a bigger review. It would be helpful if we have a sense of which models actually train (not just have some trained weights of since you'd mentioned that many models don't particularly train well)

Artifacts.toml
```diff
@@ -0,0 +1,39 @@
+[densenet121]
```
DhairyaLGandhi, line 1: Which script was used to generate these?

src/googlenet.jl
DhairyaLGandhi: Having objects shouldn't deter from being functional, and it seems like Inception might an example of a  model that would definitely benefit from a struct that holds logical components together.

src/vgg.jl
```diff
@@ -0,0 +1,248 @@
...
+                        inchannels=inchannels,
+                        nclasses=nclasses,
+                        fcsize=fcsize,
+                        dropout=dropout,
+                        batchnorm=BN)
+  
+  VGG{BN, typeof(layers)}(layers)
+end
+
+Functors.functor(::Type{VGG{BN, T}}, m) where {BN, T} =
```
DhairyaLGandhi, line 142: Do not overload it directly. Why aren't we using `@functor` here?

src/vgg.jl
```diff
@@ -0,0 +1,248 @@
...
+  (layers = m.layers,), nt -> VGG{BN, typeof(nt.layers)}(nt.layers)
+
+VGG(imsize=(224, 224); config, inchannels, nclasses, fcsize, dropout) =
```
DhairyaLGandhi, line 145:
```suggestion
VGG(imsize = (224, 224); config, inchannels, nclasses, fcsize, dropout) =
```

Where is the image size necessary knowledge? It'd be better if users didn't have to specify it at the start. Is it the fully connected layers?

DhairyaLGandhi (commented):
I would also like to test-train a few of these models to see how far we get.

Artifacts.toml
darsnack: https://github.com/FluxML/MetalheadWeights contains the scripts

src/densenet.jl
darsnack: One of the goals was to match the paper as closely as possible. So `growth_rate` corresponds to terminology in the paper (also defined as "k" in the paper which is mentioned in the docstrings).

We could combine `nblock` and `growth_rate` into `nout_channels` as a vector which I agree makes most sense. Maybe there can be a convenience method to make this vector given `nblock` and `growth_rate`.

src/densenet.jl
```diff
@@ -1,92 +1,188 @@
...
+- `reduction`: the factor by which the number of feature maps is scaled across each transition
+- `nclasses`: the number of output classes
+"""
+function densenet(nblocks; growth_rate=32, reduction=0.5, nclasses=1000)
+  num_planes = 2 * growth_rate
+  layers = []
+  append!(layers, conv_bn((7, 7), 3, num_planes; stride=2, pad=(3, 3), usebias=false))
+  push!(layers, MaxPool((3, 3), stride=2, pad=(1, 1)))
+
+  for i in 1:3
```
darsnack, line 71: This should be over `length(nblocks)`

src/densenet.jl
darsnack: That would be `DenseNet` (with no number) above. I don't think `DenseNet(121)` makes much sense cause it's not clear what `DenseNet(83)` should be. If the number argument is constrained to 121, 161, 169, and 201, then it seems cleaner to put it in the name that to take it as an argument and throw an error when it doesn't match the 5 pre-defined integers.

Once we sort out the `nblocks`/`growth_rate` issue, something like `DenseNet(...)` above seems like the clearest API for the user to specify the structure they want.

src/googlenet.jl
darsnack: Yeah, I agree it's awful. I would characterize this as an internal function (i.e. maybe we rename it to `_inceptionblock`), because the only reason we have it is to make writing `googlenet` below less verbose. It's absolutely not flexible enough to be used by users. Perhaps I could even define it inside `googlenet` so it stays private.

src/vgg.jl
darsnack: Because we need to close over the type parameters in `re`. FWIW, this is [the recommended way](https://github.com/FluxML/Functors.jl/issues/3#issuecomment-626747663) of dealing with this according to Mike.

src/vgg.jl
darsnack: Yes, the first FC layer input will be sized differently depending on your data. There's really no way around that fact for VGG. Assuming the user is using VGG for the default dataset of ImageNet, they won't have to specify the image size.

darsnack:
Thanks, I'll address most of the review comments later today.

> It would be helpful if we have a sense of which models actually train

They all actually train. But even the simplest ones (ResNet, VGG, AlexNet) are hard to train to top-k accuracy on ImageNet without using multi-GPU training. If you can share a DDP script that works, I can test it on all the models.

> (not just have some trained weights of since you'd mentioned that many models don't particularly train well)

The current pre-trained weights are the ones from `master`. I loaded them into the new models, then re-saved a BSON (which are in the artifacts repo). What I was saying is that they don't get good accuracy on the ImageNet test set currently. But then I checked out the currently released Metalhead, and the old models also don't get good accuracy.

src/densenet.jl
darsnack: It doesn't really make sense for `rev` to be user configurable here. I want to save `bias` for a later PR, so we can come up with a generic way to specify the convolution layer parameters across all model types. But I think that's an enhancement better saved for the future to prevent blocking current work in other packages. It's not like the current Metalhead allows this configuration, so it isn't a regression.

src/vgg.jl
ToucheSir: I am confused by why VGG needs a type param for batchnorm/no batchrnorm. Unlike ResNet where you'd potentially want to dispatch on the type of block, how many scenarios are there where you'd want code to behave differently depending on whether a VGG model has batchnorm?

darsnack: My thinking was to try and make the type be more than a plain wrapper, but maybe it just isn't worth it.

darsnack: So I think now I am going to forego the type parameters in `ResNet` and `VGG` since we can't provide similar types across all models, and I don't think they are that useful. In `ResNet`, it would be nice to distinguish basic block from bottleneck, but the current approach seems too hacky to me.

src/vgg.jl
DhairyaLGandhi: Hmm, maybe it's better to separate the FC layers from the rest of the model.

darsnack: The ability to specify image size here is a feature, not really something I want to remove. Why make a user calculate the FC input size when they don't need to?

darsnack:
I am considering moving a bit away from the functional form (e.g. `resnet`) to a struct. With the increased use, development, and documentation of Functors.jl, this shouldn't really be a problem. My only concern is how it affects being able to use this package as a dependency. @ablaom would it make a difference to MLJFlux if you got back a `ResNet` struct vs. a plain `Chain`?

DhairyaLGandhi:
Fwiw, we always would have returned a struct. And I think that's the better design for large models anyway.

src/resnet.jl
```diff
@@ -1,97 +1,271 @@
...
+# Arguments
+- `BC`: residual block style (either `:basic` or `:bottleneck`)
+- `SC`: shortcut style (either `:A` or `:B`)
+- `channel_config`: the growth rate of the output feature maps within a residual block
+- `block_config`: a list of the number of residual blocks at each stage
+- `nclasses`: the number of output classes
+"""
+struct ResNet{BC, SC, T}
+  layers::T
 end
```
mcabbott, line 149: With this struct outermost, I think you will want to overload `show`, behind `if isdefined(Flux, :_big_show)`. It's a nice tests case:

Screenshot 2021-07-10 at 15 18 24

ablaom:
> I am considering moving a bit away from the functional form (e.g. `resnet`) to a struct. With the increased use, development, and documentation of Functors.jl, this shouldn't really be a problem. My only concern is how it affects being able to use this package as a dependency. @ablaom would it make a difference to MLJFlux if you got back a `ResNet` struct vs. a plain `Chain`?

No, so long as I can access the chain, and this is documented.

darsnack:
Okay I think everyone's comments have been addressed. Can we please merge this? With the upcoming FastAI.jl release, it would be really nice to not need a WIP branch to use these models.

@DhairyaLGandhi Write access requests still haven't been done.

darsnack:
Any ideas on what happened with CI?

darsnack:
No clue what's wrong with the Windows CI, and nightly failures are due to BSON failing on 1.7.

DhairyaLGandhi:
Is there an issue for bson?

darsnack:
https://github.com/JuliaIO/BSON.jl/pull/102

darsnack:
All the BSON issues have been addressed in https://github.com/JuliaIO/BSON.jl/pull/102. In any case, that shouldn't be a blocker here since all the code changes are in BSON.jl.

src/Metalhead.jl
```diff
@@ -1,38 +1,33 @@
...
-# Models
-export VGG19, SqueezeNet, DenseNet, ResNet, GoogleNet
-
-# Useful re-export from Images
-export load
-
-# High-level classification APIs
-export predict, classify
+import Functors
+import Functors: functor
```
DhairyaLGandhi, line 10:
```suggestion
```

DhairyaLGandhi (approved):
Could you also add the backwards passes to be tested?

DhairyaLGandhi:
Any update on https://github.com/FluxML/Metalhead.jl/pull/70#issuecomment-877678267 ?

darsnack:
In the end I decided against it because there didn't seem like much value add for models other than VGG. For example, ResNet would just be `ResNet.backbone` and `ResNet.classifier`. To get useful information, we would have to further break down `ResNet.backbone` and "name" each block as a residual block. Introducing new types just for the sake of naming things didn't seem like a satisfying design to me. I would rather do something more generic/flexible like @ToucheSir's `NamedTuple` proposal for `Chain`s.

DhairyaLGandhi:
Well it's more for things like inception/ deeplab among other such models. The backbone would have more information too in resnet right? Plus I think we may want to have a way for users to switch out the classifier conveniently.

darsnack:
I think my response is the same for Inception as well. I agree that naming parts of a deeply nested model is useful, but I don't think that introducing new types is the right way to do it. Inception is a case where we'd end up introducing `InceptionBlockA`, `InceptionBlockB`, `InceptionBlockC`, etc., since none of the block structures are generic enough to combine into a single type.

The main issue is that the only way to index a `Chain` is with integers which are hard to keep track of. With something like the `NamedTuple` proposal, we could name the classifier element as `classifier`. Then users could do `model.classifier` to index that element. Same goes for the different inception modules.

For now, I'll wrap each major subsection as its own `Chain`. So that `myresnet[1]` gives the backbone and `myresnet[2]` gives the classifier, etc.

DhairyaLGandhi:
I think I'd prefer to have the backbone and classifier separated since there are many models that drop the classifier to use resnet as a backbone, so this should make it easier.

Any reason why we couldn't achieve the same results of "indexing" using structs. It seems much more Julian.

darsnack:
It would still be easy to drop the classifier. `model = Chain(resnet.backbone, myclassifier)` vs `model = Chain(resnet[1], myclassifier)`.

I would argue the struct approach is not Julian and more Python. Defining tons of little classes or wrapper classes is very PyTorch-like. You achieve the same "indexing" result, but if all you need is named indexing, then I don't see why you wouldn't use a `NamedTuple`-like solution. If you introduce a new struct type just for naming, it comes at a higher cost. You have boilerplate to `@functor` the type, and you need to define the forward pass. In the current approach we need to do neither. We take advantage of the higher level layers available in Flux to avoid doing any of that boiler plate.

More importantly, going with the struct solution means we only solve the naming problem for these models in this repo. It isn't a flexible solution that helps anyone else. I'd say that on top of anything is the best reason to solve this more generically.

DhairyaLGandhi:
Well if we see Chains everywhere anyway, its harder to tell what exactly is happening in the forwards pass, and named tuples aren't great for when you have large objects inside them. structs are good for containerisation and labelling. Forwards pass add some clarity towards the intent of what the layer is doing (unless its `l.model(x)` of course :)). I agree we should have a nicer solution to avoid the boilerplate, which is where the functional aspect is better. Not a fan of boilerplate here :)

DhairyaLGandhi:
Its good that Functors can already handle most of the plumbing we would want from these models

DhairyaLGandhi:
We might need to ease up on the batch size for CI. Could we make it 2 or something?

darsnack:
Is there a performance degradation when using `NamedTuple`s vs.` Tuple`s for large objects? `Chain` already uses `Tuple`s. TBH the biggest difference w.r.t. knowing what the forward pass does has been @mcabbott's "big show" method. Without that, it wouldn't really matter what the internal representation was cause it would just print like a big blob.

I'm still a bit unconvinced about this issue. Can we solve this issue in another PR to Metalhead or Flux (depending on which solution we select)? Right now, the outermost object is a struct. It's possible to swap out parts of the model easily like I described above. We can always enrich the repo later instead of bikeshedding on this issue now.

DhairyaLGandhi:
Wouldn't you want a separate show method for different containers? Looking at Chains is the bottleneck to me. I'd love to see this in once CI is happy.

We should try to bring in bindings for datasets etc back and have docs using FluxML - we can do that later.

DhairyaLGandhi (commented):
It may be too much for GA to handle large models on CI - we'll set up a buildkite but in the meanwhile.

DhairyaLGandhi:
Thanks all! I'll let this be on master for a couple days before releasing.

lorenzoh:
Great! I'll update the install instructions in FastAI.jl as soon as a new release is tagged.

ablaom:
Any chance we could a new release tagged? This was merged quite a while ago.

DhairyaLGandhi:
see FluxML/FluxBench.jl#9


I also want to know the status of the training pipeline for pretrained weights as a todo. I plan on releasing this this week or next.

DhairyaLGandhi:
We also need to get rid of some unnecessary typing for example https://github.com/FluxML/Metalhead.jl/blob/3f458cd0db47a630a09ac5fd1bad2beaeb549941/src/squeezenet.jl#L60 and the like, since that leads to an extended compile time without offering runtime benefits.

darsnack:
So would you want to just leave them untyped?

DhairyaLGandhi:
I think it would be fine to have some kind of helpful hints that are useful for dispatch, but avoid a massive type when its not necessary for inference. Currently, this would yield a massive type, which slows down the compiler and type inference significantly. Since the layers themselves are typed already, we get the inference either way. In fact, we may want to consider dropping the type param from Chain and checking the impact of that on small models. Both latency and runtime wise.

ToucheSir:
Can't find my earlier comment on this, but could we define type aliases for certain basic blocks in order to help with display? AIUI that was a big reason for having dedicated types in the first place, right?

ablaom:
Are we waiting for some more work before tagging a new release?

DhairyaLGandhi:
Yeah, there's a few cleanups that we need before a full release. I'll get a pr in a couple days.

PR #71, opened 2020-09-07
add inceptions

DhairyaLGandhi:
These are some of the inceptions that I have had hanging around for a while, which should be a drop in replacement to the newer Flux versions. Need to update the `functor` calls and use more flexible APIs to generate these.


darsnack:
Closed by #170

PR #75, opened 2021-01-24
Bump compat of Colors

clad26:


DhairyaLGandhi:
Flux v0.12 isn't released yet, so it's a bit weird claiming compat with it.

clad26:
Already mentioned this in the slack, but anyway

> @Dhairya Yeah! Its the master branch of Flux
The pull which fixes Flux's Project.toml version problem with Adapt.jl will get merged to the master branch of Flux, which is 0.12.0dev, and won't affect 0.11.4 in any way. So even if both get merged, it doesn't solve the problem with Flux 0.11.4. The problem only gets solved with 0.12.0dev. My suggestion is, merge the Metalhead PR, and release a new minor version of Metalhead when Flux 0.12.0 is released?

clad26:
@DhairyaLGandhi your suggestion on the above opinion is appreciated!

DhairyaLGandhi:
Well, we can break things on Flux master still, so :shrug: 

> release a new minor version of Metalhead when Flux 0.12.0 is released?

Yes, but we could also bump colortypes, register a release and have the flux compat only on master via the manifest.

If that sounds alright, you could update the PR accordingly and bump the patch version, I can tag a release of Metalhead.

clad26:
@DhairyaLGandhi You're releasing Flux 0.11.5?
In that case, I can remove the Flux 0.12 dep from the PR!

clad26:
Ready for release, I guess?😃

PR #91, opened 2021-10-07
Make ResNet more flexible/generic + minor bug fix

darsnack:
This PR does not functionally change any of the ResNet variants provided. It adds a new, more flexible and more intuitive, interface for the builder `resnet`. It also fixes a bug with shortcut style C. Below is a summary of the changes.

- `resnet(block, residuals, connection; kwargs...)` is the new underlying builder interface. This lets a user pass in any function for the `residuals` and `connection` (previously this was hardcoded to `skip_identity`/`skip_projection` based on the more opaque keyword `shortcut_config`).
- For users familiar with the ResNet papers, the old interface for specifying a `shortcut_config` as `:A`, `:B`, or `:C` still exists. I added more details to the docstring to make this keyword less opaque.
- There was an inconsistency in the arguments for `skip_identity` and `skip_projection` that cause an error when `shortcut_config == :C`. Previously this was not caught because none of the default variants use this shortcut style even though it is mentioned in the paper. Now there is a separate test for this shortcut style, and the original bug/inconsistency is fixed.
- The utility `conv_bn` now accepts all the arguments and keywords that `Conv` and `BatchNorm` do (previously this only included `stride` and `pad`). It also accepts an activation function argument (previously hardcoded to `relu`).
- `resnet` takes advantage of the change to `conv_bn` to push the final activation for each residual block into the `connection`. This is functionally the same as before, but it allows users to pass in `connection = (x, y) -> @. relu(x + y)`. This style of connection wasn't in the original paper but is commonly found in implementations of ResNet in various frameworks.

### DIFF
--- a/src/resnet.jl
+++ b/src/resnet.jl
@@ -119,8 +119,8 @@ function resnet(block, residuals::NTuple{2, Any}, connection = (x, y) -> @. relu
     baseplanes *= 2
   end
 
-  return Chain(Chain(layers..., AdaptiveMeanPool((1, 1))),
-               Chain(flatten, Dense(inplanes, nclasses)))
+  return Chain(Chain(layers...),
+               Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
 end
 
 """


darsnack:
I have one more PR (incoming in a sec) with a couple of final tweaks to the other models, but then I think we should release.

src/resnet.jl
```diff
@@ -44,83 +44,111 @@ Create a skip projection
...
     for _ in 2:nrepeats
-      if shortcut_config == :A || shortcut_config == :B
-        push!(layers, Parallel(+, block(inplanes, outplanes, false),
-                                  skip_identity(inplanes, outplanes[end])))
-      elseif shortcut_config == :C
-        push!(layers, Parallel(+, block(inplanes, outplanes, false),
-                                  skip_projection(inplanes, outplanes, false)))
-      end
+      push!(layers, Parallel(connection, block(inplanes, outplanes, false),
+                                         residuals[1](inplanes, outplanes[end], false)))
```
iblis17, line 115: Should this be `residuals[2]`, according to the comment?

darsnack: Yes! Thank you for catching this. Would you like to submit a PR to fix this? If not, I will later today.

iblis17: Ah, sure. https://github.com/FluxML/Metalhead.jl/pull/97

PR #92, opened 2021-10-09
Add references/links to models in README.md

trahflow:
Add references/papers to model classes and links to the implementation in this package

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,28 +14,28 @@
 
 ## Available models
 
-| Model Name     | Function           | Pre-trained? |
-|:---------------|:-------------------|:------------:|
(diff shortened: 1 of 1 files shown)


darsnack:
The old version linked to the docstrings for each model when viewed as the webpage. If we want the links to work on GitHub too, we can manually add links to the docstrings instead of the source code.

trahflow:
> we can manually add links to the docstrings instead of the source code.
Not sure how to achieve this (other than referencing a line number) with markdown in github

darsnack:
We should make sure the docs CI is running and put a badge on the README.

PR #93, opened 2021-10-10
Release fixes

darsnack:
Some last minute fixes before releasing:
- Adaptive pooling layers are moved to classifier section instead of backbone
- Disabled the pre-trained models since we determined that the old weights are not producing accurate results; I don't think we should ship broken models
- Fix CI badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Metalhead
 
+[![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://fluxml.github.io/Metalhead.jl/dev)
 [![CI](https://github.com/FluxML/Metalhead.jl/actions/workflows/CI.yml/badge.svg)](https://github.com/FluxML/Metalhead.jl/actions/workflows/CI.yml)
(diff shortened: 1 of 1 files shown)


darsnack:
@lorenzoh will removing the pre-trained models cause issues for FastAI.jl? They shouldn't have been producing accurate results if you were using them.

darsnack:
Training even the subset of models that we had before would take some time. Would you prefer to ship with the same pre-trained models but potentially inaccurate behavior?

DhairyaLGandhi:
Yeah, I noticed the models behavior, agree we shouldn't be shipping them yet. I have some hardware that I'm waiting on that's earmarked for exactly this.

there's also a few other things, such as dropping the type params on the large models to avoid over specialisation, and a uniform API to separate access the backbone and relax the requirement to give the sizes of the inputs to create the models

DhairyaLGandhi:
> Something like backbone(model) and classifier(model) okay?

Ideally the models would be the backbone themselves and classifiers can be added on top. That way our API surface is also low.

darsnack:
But a ResNet50 specifies both a backbone and a classifier. If I write `model = ResNet50()` in my code and I get a model that doesn't include the classifier that is confusing. For research cases where reproducibility matters, it's important for us to return the exact model with easy ways to customize.

PR #99, opened 2022-01-10
Simplify construction for some models

DhairyaLGandhi:
Moves us from having multiple functions such as `ResNet18/34/50` etc to a unified constructor which can accept the `depth` and construct the corresponding model. It simplifies the code base by removing some stencil functions and allows us to get a default `ResNet()` method as well for backwards compatibility with Metalhead@0.5.

### DIFF
--- a/src/resnet.jl
+++ b/src/resnet.jl
@@ -214,7 +214,7 @@ See also [`Metalhead.resnet`](#).
     Only `ResNet(50)` currently supports pretrained weights.
 """
 function ResNet(depth::Int = 50; pretrain = false, nclasses = 1000)
-    @assert depth in (18, 34, 50, 101, 152) "`depth` must be one of (18, 34, 50, 101, 152)"
+    @assert depth in keys(resnet_config) "`depth` must be one of $(sort(collect(keys(resnet_config))))"
 
     config, block = resnet_config[depth]
     model = ResNet(config...; block = block, nclasses = nclasses)
--- a/src/resnet.jl
+++ b/src/resnet.jl
@@ -211,3 +211,10 @@ function ResNet(depth = 50; pretrain = false, nclasses = 1000)
     pretrain && loadpretrain!(model, string("ResNet", config))
     model
 end
+
+# Compat with Methalhead 0.6; remove in 0.7
+@deprecate ResNet18() ResNet(18)
+@deprecate ResNet34() ResNet(34)
+@deprecate ResNet50() ResNet(50)
+@deprecate ResNet101() ResNet(101)
+@deprecate ResNet152() ResNet(152)
--- a/src/resnet.jl
+++ b/src/resnet.jl
@@ -29,7 +29,7 @@ Create a bottleneck residual block
 - `downsample`: set to `true` to downsample the input
 """
 function bottleneck(inplanes, outplanes, downsample = false)
-  stride = downsample : 2 : 1
+  stride = downsample ? 2 : 1
   Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = stride, bias = false)...,
         conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 1, pad = 1, bias = false)...,
         conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...)
--- a/src/resnet.jl
+++ b/src/resnet.jl
@@ -143,18 +143,25 @@ Create a ResNet model
 - `block_config`: a list of the number of residual blocks at each stage
 - `nclasses`: the number of output classes
 """
-resnet(block, shortcut_config::Symbol, args...; kwargs...) =
-  (shortcut_config == :A) ? resnet(block, (skip_identity, skip_identity), args...; kwargs...) :
-  (shortcut_config == :B) ? resnet(block, (skip_projection, skip_identity), args...; kwargs...) :
-  (shortcut_config == :C) ? resnet(block, (skip_projection, skip_projection), args...; kwargs...) :
-  error("Unrecognized shortcut config == $shortcut_config passed to resnet (use :A, :B, or :C).")
+function resnet(block, shortcut_config::Symbol, args...; kwargs...)
+  shortcut = if shortcut_config == :A
+      (skip_identity, skip_identity)
+    elseif shortcut_config == :B
+      (skip_projection, skip_identity)
+    elseif shortcut_config == :C
+      (skip_projection, skip_projection)
+    else
+      error("Unrecognized shortcut_config ($shortcut_config) passed to `resnet` (use :A, :B, or :C).")
+  end
+  resnet(block, shortcut, args...; kwargs...)
+end
 
 const resnet_config =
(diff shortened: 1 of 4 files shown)


darsnack:
The reason I chose not to go this route is that it makes depth seem like you could give it any value when in reality it only supports a fixed subset defined in the paper. Contrast this with the multi-arg `ResNet` that already exists which can support any coherent set of values.

DhairyaLGandhi:
We are able to say this through the documentation, and it helps keep the code terse. I understand why this choice was made, but in reality, it is acting mostly as a simple convenience constructor so I hope it will be fine to relax the API.

darsnack:
There should at least be a check in the new constructor and a more informative error then.

I'm still unconvinced that this is better though. What's the use case for this? Not trying to block the change, just trying to understand why.

DhairyaLGandhi:
Having a simpler code base is better for making sure we don't have to make the same changes several times in the repo, and the API is high enough level that it is unlikely that it would result in confusion. It is clearly documented that we are referring to the implementation from the paper as well.

src/resnet.jl
```diff
@@ -193,105 +200,31 @@ backbone(m::ResNet) = m.layers[1]
...
 - `nclasses`: the number of output classes
 
 !!! warning
-    `ResNet18` does not currently support pretrained weights.
+    Only `ResNet(50)` currently supports pretrained weights.
 """
-function ResNet18(; pretrain = false, nclasses = 1000)
-  model = ResNet(resnet_config[:resnet18]...; block = basicblock, nclasses = nclasses)
+function ResNet(depth::Int = 50; pretrain = false, nclasses = 1000)
+    @assert depth in (18, 34, 50, 101, 152) "`depth` must be one of (18, 34, 50, 101, 152)"
```
DhairyaLGandhi, line 217: @darsnack This lists acceptable depth values. Anything we need to clarify here? Happy to hear your thoughts on what would be a better message.

darsnack: Ah, I'm sorry I missed this! Looks good enough for me.

ToucheSir: Is it possible to reuse `keys(resnet_config)` for this or is that not guaranteed to be ordered?

darsnack: `sort(keys(resnet_config))` I guess

DhairyaLGandhi: Done.

src/resnet.jl
```diff
@@ -10,11 +10,11 @@ Create a basic residual block
...
 """
-basicblock(inplanes, outplanes, downsample = false) = downsample ?
-  Chain(conv_bn((3, 3), inplanes, outplanes[1]; stride = 2, pad = 1, bias = false)...,
-        conv_bn((3, 3), outplanes[1], outplanes[2], identity; stride = 1, pad = 1, bias = false)...) :
-  Chain(conv_bn((3, 3), inplanes, outplanes[1]; stride = 1, pad = 1, bias = false)...,
+function basicblock(inplanes, outplanes, downsample = false)
+  stride = downsample ? 2 : 1
+  Chain(conv_bn((3, 3), inplanes, outplanes[1]; stride = stride, pad = 1, bias = false)...,
         conv_bn((3, 3), outplanes[1], outplanes[2], identity; stride = 1, pad = 1, bias = false)...)
+end
```
darsnack, line 17: I'm not sure this change is equivalent since in the `downsample == false` case, there should only be 1 conv.

src/resnet.jl
```diff
@@ -28,13 +28,12 @@ Create a bottleneck residual block
...
-  Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = 2, bias = false)...,
-        conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 1, pad = 1, bias = false)...,
-        conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...) :
-  Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = 1, bias = false)...,
+function bottleneck(inplanes, outplanes, downsample = false)
+  stride = downsample ? 2 : 1
+  Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = stride, bias = false)...,
         conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 1, pad = 1, bias = false)...,
         conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...)
+end
```
darsnack, line 36: Same comment here

src/vgg.jl
```diff
@@ -144,117 +145,38 @@ backbone(m::VGG) = m.layers[1]
...
-    `VGG16` does not currently support pretrained weights.
+function VGG(depth::Int = 16; pretrain = false, batchnorm = false)
+  @assert depth in keys(vgg_config) "depth must be from one in $(sort(collect(keys(vgg_config))))"
 
-# Arguments
-- `pretrain`: set to `true` to load pre-trained model weights for ImageNet
-"""
-function VGG16(; pretrain = false, batchnorm = false)
-  model = VGG((224, 224); config = vgg_config[:D],
+  model = VGG((224, 224); config = vgg_conv_config[vgg_config[depth]],
```
darsnack, line 163: Any reason for the double dict?

src/vgg.jl
```diff
@@ -97,41 +97,42 @@ function vgg(imsize; config, inchannels, batchnorm = false, nclasses, fcsize, dr
...
+                        16 => :D,
+                        19 => :E)
+
+struct VGG
+  layers
+end
 
 """
-    VGG(imsize = (224, 224); config, inchannels, batchnorm = false, nclasses, fcsize, dropout)
+    VGG(imsize::NTuple{2,Int}; config, inchannels, batchnorm, nclasses, fcsize, dropout)
```
darsnack, line 115:
```suggestion
    VGG(imsize::NTuple{2,Int}; config, inchannels, batchnorm = false, nclasses, fcsize, dropout)
```

src/vgg.jl
```diff
@@ -144,117 +145,38 @@ backbone(m::VGG) = m.layers[1]
...
-"""
-    VGG16(; pretrain = false, batchnorm = false)
-
-Create a VGG-11 style model
-([reference](https://arxiv.org/abs/1409.1556v6)).
-See also [`VGG`](#).
-
-!!! warning
-    `VGG16` does not currently support pretrained weights.
+function VGG(depth::Int = 16; pretrain = false, batchnorm = false)
```
jeremiedb, line 160: Should add a `nclasses` argument to mirror ResNet's options?

DhairyaLGandhi: Good idea, I was also thinking of a `bias` option (perhaps in a later PR) to switch bias from user input.

src/vgg.jl
DhairyaLGandhi: Mostly to mirror the other models and to retain the same nomenclature of the model config as the paper. I was considering changing the config such that we go `11 => A => ...` too.

darsnack (approved):

PR #100, opened 2022-01-15
Implementation of ResNeXt

theabhirath:
This code implements [ResNeXt](https://arxiv.org/pdf/1611.05431.pdf), a convolutional neural network that is often used for image classification and as backbone for image segmentation tasks.

### DIFF
--- a/src/resnext.jl
+++ b/src/resnext.jl
@@ -11,14 +11,16 @@ Create a basic residual block as defined in the paper for ResNeXt
 - `width`: the number of feature maps in each group in the bottleneck
 - `downsample`: set to `true` to downsample the input
 """
-function resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
+function resnextblock(inplanes, outplanes, cardinality, width, widen_factor, downsample = false)
     stride = downsample ? 2 : 1
-    inner_channels = cardinality * width
 
-    return Chain(conv_bn((1, 1), inplanes, inner_channels; stride = 1, bias = false)...,
-        conv_bn((3, 3), inner_channels, inner_channels;
+    width_ratio = outplanes / (widen_factor * 128.0)
+    hidden_channels = cardinality * floor(width * width_ratio)
+
+    return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
+        conv_bn((3, 3), hidden_channels, hidden_channels;
             stride = stride, pad = 1, bias = false, groups = cardinality)...,
-        conv_bn((1, 1), inner_channels, outplanes; stride = 1, bias = false)...)
+        conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
 end
 
 """
@@ -35,7 +37,7 @@ Create a ResNeXt model
 - `block_config`: a list of the number of residual blocks at each stage
 - `nclasses`: the number of output classes
 """
-function resnext(cardinality, width, channel_multiplier = 2, connection = (x, y) -> @. relu(x) + relu(y);
+function resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @. relu(x) + relu(y);
     block_config, nclasses = 1000)
     inplanes = 64
     baseplanes = 128
@@ -44,15 +46,15 @@ function resnext(cardinality, width, channel_multiplier = 2, connection = (x, y)
     push!(layers, MaxPool((3, 3), stride = (2, 2), pad = (1, 1)))
     for (i, nrepeats) in enumerate(block_config)
         # output planes within a block
-        outplanes = baseplanes * channel_multiplier
+        outplanes = baseplanes * widen_factor
         # push first skip connection on using first residual
         # downsample the residual path if this is the first repetition of a block
-        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
+        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, i != 1),
             skip_projection(inplanes, outplanes, i != 1)))
         # push remaining skip connections on using second residual
         inplanes = outplanes
         for _ in 2:nrepeats
-            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, false),
+            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, false),
                 skip_identity(inplanes, outplanes, false)))
         end
         baseplanes = outplanes
--- a/src/resnext.jl
+++ b/src/resnext.jl
@@ -37,32 +37,31 @@ Create a ResNeXt model
 - `cardinality`: the number of groups to use for the convolution
 - `nclasses`: the number of output classes
 """
-function resnext(connection = (x, y) -> @. relu(x) + relu(y); channel_config, block_config, cardinality, nclasses = 1000)
-    inplanes = 128
+function resnext(cardinality, width, channel_multiplier = 2, connection = (x, y) -> @. relu(x) + relu(y);
+                 block_config, nclasses = 1000)
+    inplanes = 64
     baseplanes = 128
     layers = []
     append!(layers, conv_bn((7, 7), 3, inplanes; stride = 2, pad = (3, 3)))
     push!(layers, MaxPool((3, 3), stride = (2, 2), pad = (1, 1)))
     for (i, nrepeats) in enumerate(block_config)
         # output planes within a block
-        outplanes = baseplanes .* channel_config
+        outplanes = baseplanes * channel_multiplier
         # push first skip connection on using first residual
         # downsample the residual path if this is the first repetition of a block
-        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, i != 1),
-                                            skip_projection(inplanes, outplanes[end], i != 1)))
+        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
+                                           skip_projection(inplanes, outplanes, i != 1)))
         # push remaining skip connections on using second residual
-        inplanes = outplanes[end]
+        inplanes = outplanes
         for _ in 2:nrepeats
-            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, false),
-                                                skip_identity(inplanes, outplanes[end], false)))
-            inplanes = outplanes[end]
+            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, false),
+                                               skip_identity(inplanes, outplanes, false)))
         end
-        # next set of output plane base is doubled
-        baseplanes *= 2
+        baseplanes = outplanes
     end
 
     return Chain(Chain(layers...),
-                Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
+                 Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
 end
 
 """
--- a/src/resnext.jl
+++ b/src/resnext.jl
@@ -11,11 +11,10 @@ Create a basic residual block as defined in the paper for ResNeXt
 - `width`: the number of feature maps in each group in the bottleneck
 - `downsample`: set to `true` to downsample the input
 """
-function resnextblock(inplanes, outplanes, cardinality, width, widen_factor, downsample = false)
+function resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
     stride = downsample ? 2 : 1
 
-    width_ratio = outplanes / (widen_factor * 128.0)
-    hidden_channels = cardinality * floor(width * width_ratio)
+    hidden_channels = cardinality * width
 
     return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
         conv_bn((3, 3), hidden_channels, hidden_channels;
@@ -38,7 +37,7 @@ Create a ResNeXt model
 - `nclasses`: the number of output classes
 """
 function resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @. relu(x) + relu(y);
-    block_config, nclasses = 1000)
+                block_config, nclasses = 1000)
     inplanes = 64
     baseplanes = 128
     layers = []
@@ -49,15 +48,16 @@ function resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @.
         outplanes = baseplanes * widen_factor
         # push first skip connection on using first residual
         # downsample the residual path if this is the first repetition of a block
-        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, i != 1),
+        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
             skip_projection(inplanes, outplanes, i != 1)))
         # push remaining skip connections on using second residual
         inplanes = outplanes
         for _ in 2:nrepeats
-            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, false),
+            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, false),
                 skip_identity(inplanes, outplanes, false)))
         end
         baseplanes = outplanes
+        width *= widen_factor
     end
 
     return Chain(Chain(layers...),
@@ -93,11 +93,12 @@ classifier(m::ResNeXt) = m.layers[2]
 
 const resnext_config = Dict(
     50 => (3, 4, 6, 3),
-    101 => (3, 4, 23, 3)
+    101 => (3, 4, 23, 3),
+    152 => (3, 8, 36, 3)
 )
 
 """
-    ResNeXt(; config::Int = 50, cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
+    ResNeXt(config::Int = 50; cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
 
 Create a ResNeXt model with specified configuration. Currently supported values are (50, 101).
 ([reference](https://arxiv.org/abs/1611.05431)).
@@ -108,7 +109,7 @@ Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
 
 See also [`Metalhead.resnext`](#).
 """
-function ResNeXt(; config::Int = 50, cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
+function ResNeXt(config::Int = 50; cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
     @assert config in keys(resnext_config) "`config` must be one of $(sort(collect(keys(resnext_config))))"
 
     model = ResNeXt(cardinality, width; block_config = resnext_config[config], nclasses)
--- a/src/resnext.jl
+++ b/src/resnext.jl
@@ -63,7 +63,7 @@ function resnext(cardinality, width, channel_multiplier = 2, connection = (x, y)
 end
 
 """
-    ResneXt(cardinality, width; block_config, nclasses = 1000)
+    ResNeXt(cardinality, width; block_config, nclasses = 1000)
     
 Create a ResNeXt model
 ([reference](https://arxiv.org/abs/1611.05431)).
--- a/src/resnext.jl
+++ b/src/resnext.jl
@@ -1,13 +1,16 @@
-using Flux
-using Functors
-using Flux: Zygote
-using Flux: _big_show
-using Metalhead
-using Test
-using Metalhead: conv_bn
-using Metalhead: skip_identity
-using Metalhead: skip_projection
+"""
+    resnextblock(inplanes, outplanes, cardinality, downsample = false)
 
+Create a basic residual block as defined in the paper for ResNeXt.
+([reference](https://arxiv.org/abs/1611.05431)).
+
+# Arguments:
+- `inplanes`: the number of input feature maps
+- `outplanes`: a list of the number of output feature maps for each convolution
+               within the residual block
+- `cardinality`: the number of groups to use for the convolution
+- `downsample`: set to `true` to downsample the input
+"""
 resnextblock(inplanes, outplanes, cardinality, downsample = false) = downsample ? 
     Chain(
         conv_bn((1, 1), inplanes, outplanes[1]; stride = 1, bias = false)...,
@@ -20,7 +23,20 @@ resnextblock(inplanes, outplanes, cardinality, downsample = false) = downsample 
                     groups = cardinality, bias = false)...,
         conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...)
 
+"""
+    resnext(connection = (x, y) -> @. relu(x) + relu(y); channel_config, block_config, 
+            cardinality, nclasses = 1000)
+    
+Create a ResNeXt model
+([reference](https://arxiv.org/abs/1611.05431)).
 
+# Arguments
+- `connection`: the binary function applied to the output of residual and skip paths in a block
+- `channel_config`: the growth rate of the output feature maps within a residual block
+- `block_config`: a list of the number of residual blocks at each stage
+- `cardinality`: the number of groups to use for the convolution
+- `nclasses`: the number of output classes
+"""
 function resnext(connection = (x, y) -> @. relu(x) + relu(y); channel_config, block_config, cardinality, nclasses = 1000)
     inplanes = 128
     baseplanes = 128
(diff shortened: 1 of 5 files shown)


theabhirath:
The `show` function doesn't seem to be working as expected for large models on this repo, however.

Screenshot 2022-01-15 at 8 13 37 PM

theabhirath:
Resolved previous issues, this now seems good to go.

darsnack:
I will need to look through the paper later tonight to make sure everything is good before merging, but on a first pass, this looks great.

The only changes for now are to fix minor formatting issues like code lines not being lined up. Also, can you update the constructors for ResNeXt50 etc. to match #99?

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+([reference](https://arxiv.org/abs/1611.05431)).
+
+# Arguments
+- `connection`: the binary function applied to the output of residual and skip paths in a block
+- `channel_config`: the growth rate of the output feature maps within a residual block
+- `block_config`: a list of the number of residual blocks at each stage
+- `cardinality`: the number of groups to use for the convolution
+- `nclasses`: the number of output classes
+"""
+function resnext(connection = (x, y) -> @. relu(x) + relu(y); channel_config, block_config, cardinality, nclasses = 1000)
```
DhairyaLGandhi, line 40:
```suggestion
function resnext(connection = (x, y) -> @. relu(x) + relu(y);
                             channel_config, block_config, cardinality, nclasses = 1000)
```

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+
+!!! warning
+    `ResNeXt50` does not currently support pretrained weights.
+"""
+function ResNeXt50(; pretrain = false, nclasses = 1000)
+    channel_config = [1, 1, 2]
+    block_config = [3, 4, 6, 3]
+    cardinality = 32
+    return ResNeXt(channel_config, block_config, cardinality; nclasses)
+end
```
DhairyaLGandhi, line 117: See #99 for the simpler API

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+"""
+struct ResNeXt
+    layers
+end
+
+function ResNeXt(channel_config, block_config, cardinality; nclasses = 1000)
+    layers = resnext(;channel_config = channel_config,
+                  block_config = block_config,
+                  cardinality = cardinality,
+                  nclasses = nclasses)
```
DhairyaLGandhi, line 88: Seems to me that the API is backwards here. `channel_config`, `block_config` and `cardinality` need to be positional, and have convenience constructors for skipping the `connection`.

DhairyaLGandhi (commented):
This is looking great so far! Could you also add this to the tests please? Looking forward to the transformers!

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+    Chain(
+        conv_bn((1, 1), inplanes, outplanes[1]; stride = 1, bias = false)...,
+        conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 2, pad = 1, 
+                    groups = cardinality, bias = false)...,
+        conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...) :
+    Chain(
+        conv_bn((1, 1), inplanes, outplanes[1]; stride = 1, bias = false)...,
+        conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 1, pad = 1,
+                    groups = cardinality, bias = false)...,
+        conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...)
```
darsnack, line 24: I believe the ResNeXt block is best understood as # of in channels, cardinality, and bottleneck width (see Fig. 3c in the paper). As such, the API for the block should probably be more like the following:
```julia
function resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
    stride = downsample ? 2 : 1
    inner_channels = cardinality * width

    return Chain(conv_bn((1, 1), inplanes, inner_channels; stride = 1, bias = false)...,
                 conv_bn((3, 3), inner_channels, inner_channels;
                         stride = stride, pad = 1, bias = false, groups = cardinality)...,
                 conv_bn((1, 1), inner_channels, outplanes; stride = 1, bias = false)...)
end
```

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+                                                skip_identity(inplanes, outplanes[end], false)))
+            inplanes = outplanes[end]
+        end
+        # next set of output plane base is doubled
+        baseplanes *= 2
+    end
+
+    return Chain(Chain(layers...),
+                Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
+end
```
darsnack, line 66: As a result of my comment about `resnextblock`, we don't really need `channel_config` here. This is one of the features of ResNeXt...it simplifies the design quite a bit by thinking more generically. In the paper, `channel_config == [1, 1, 2]` always, so we could just hardcode the fact that the output channels doubles at the end of the block. But I think future-proof this a bit by adding a `channel_multiplier`. Here's how I would do it.

Also, note a bug that `inplanes = 64` to start.

```suggestion
function resnext(cardinality, width, channel_multiplier = 2, connection = (x, y) -> @. relu(x) + relu(y);
                 block_config, nclasses = 1000)
    inplanes = 64
    baseplanes = 128
    layers = []
    append!(layers, conv_bn((7, 7), 3, inplanes; stride = 2, pad = (3, 3)))
    push!(layers, MaxPool((3, 3), stride = (2, 2), pad = (1, 1)))
    for (i, nrepeats) in enumerate(block_config)
        # output planes within a block
        outplanes = baseplanes * channel_multiplier
        # push first skip connection on using first residual
        # downsample the residual path if this is the first repetition of a block
        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
                                           skip_projection(inplanes, outplanes, i != 1)))
        # push remaining skip connections on using second residual
        inplanes = outplanes
        for _ in 2:nrepeats
            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, false),
                                               skip_identity(inplanes, outplanes, false)))
        end
        baseplanes = outplanes
    end

    return Chain(Chain(layers...),
                 Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
end
```

src/resnext.jl
darsnack: That's because this mirrors the `resnet` API (as it should for consistency) and the `resnet` API contains additional positional arguments that are more important, but those arguments are dropped here. In any case, I think the API needs some more significant changes since ResNeXt is specified in terms of cardinality and width, not a channel configuration. I've suggested an alternate API where the relevant positional args are given first.

src/resnext.jl
```diff
@@ -0,0 +1,137 @@
...
+- `channel_config`: the growth rate of the output feature maps within a residual block
+- `block_config`: a list of the number of residual blocks at each stage
+- `cardinality`: the number of groups to use for the convolution
+- `nclasses`: the number of output classes
+"""
+struct ResNeXt
+    layers
+end
+
+function ResNeXt(channel_config, block_config, cardinality; nclasses = 1000)
```
darsnack, line 84: Accordingly, this should be `ResNeXt(block_config, cardinality, width; nclasses = 1000)`.

darsnack (changes requested):
Despite the big diffs here, I think the initial implementation is solid. The major changes are to the API which should be based primarily on cardinality and bottleneck width, since that is how various ResNeXt configurations are formed.

Also, +1 on @DhairyaLGandhi's suggestions for simplifying the constructors and adding tests.

Thanks for this PR, it will make a great addition to the package!

src/resnext.jl
DhairyaLGandhi: I'd rather avoid hard coding it if there are variants of interest that diverge from the default. It keeps the design simpler by hoisting the `channel_multiplier` outside the package. If the dimensions don't align, that would error on the forward pass, so we should be fine to report those.

darsnack: Not sure what you mean exactly, but my suggestion is not to hardcode it.

src/resnext.jl
theabhirath: Yeah, I thought of doing away with the original ResNet style API but this is my first contribution so was a little unsure of how much I could change. The cardinality x width specification is much more characteristic for ResNeXt, I agree.

theabhirath:
Okay so while writing the tests I noticed they're still in the style of the old API for the ResNet, so should I take the liberty of updating it so that the ResNet and ResNeXt tests match the new API as proposed in #99 ?

theabhirath:
There is also the addition problem that since the struct and the function to integrate with it as well as the user API have the same name, VSCode IntelliSense seems to be messing up the display of documentation on hover - it's showing both of them back-to-back which is not ideal. Simply changing the function name to have a small n (the struct and the function to integrate with it) fixes this, so if that's okay I can make that change as well.

darsnack:
> Okay so while writing the tests I noticed they're still in the style of the old API for the ResNet, so should I take the liberty of updating it so that the ResNet and ResNeXt tests match the new API as proposed in #99 ?

Yes, just for ResNeXt. The other models should stay as is, since they will be handled in #99.

> VSCode IntelliSense seems to be messing up the display of documentation on hover - it's showing both of them back-to-back

Are you referring to the docstring for `ResNeXt` with the new API? It will show both the docstring above the struct as well as above the convenience constructor for pre-trained models? This is not a problem. In Julia, it is typical for a single function (i.e. `ResNeXt`) to have multiple methods (i.e. the default constructor and the convenience constructor). When a user types `ResNeXt` in their code, they may want either constructor which is why the VSCode extension displays both. Sometimes, we merge all of these definitions into a single docstring above the struct, but if they are different enough (as is the case here), then we would prefer separate docstrings. I would not change one to have a lower case "n" and the other capitalized (maybe I misunderstood the suggestions, so feel free to clarify).

src/resnext.jl
```diff
@@ -0,0 +1,117 @@
...
+    stride = downsample ? 2 : 1
+
+    width_ratio = outplanes / (widen_factor * 128.0)
+    hidden_channels = cardinality * floor(width * width_ratio)
+
+    return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
+        conv_bn((3, 3), hidden_channels, hidden_channels;
+            stride = stride, pad = 1, bias = false, groups = cardinality)...,
+        conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
+end
```
darsnack, line 24: I don't believe the `widen_factor` should be considered here. The inner channel dimension is always `cardinality * width`, and it is independent of the input and output channels. See Fig. 3c in the paper: the input and output channels are 256, and the inner channels is 32*4 = 128.

src/resnext.jl
```diff
@@ -0,0 +1,117 @@
...
+
+# Arguments
+- `cardinality`: the number of groups to use for the convolution
+- `width`: the number of feature maps in each group in the bottleneck
+- `connection`: the binary function applied to the output of residual and skip paths in a block
+- `block_config`: a list of the number of residual blocks at each stage
+- `nclasses`: the number of output classes
+"""
+function resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @. relu(x) + relu(y);
+    block_config, nclasses = 1000)
```
darsnack, line 41: Make sure to follow the style of the rest of the code base. We don't indent by only 1 on a new line. `block_config` should line up with `cardinality` here.

src/resnext.jl
```diff
@@ -0,0 +1,117 @@
...
+Create a ResNeXt model with specified configuration. Currently supported values are (50, 101).
+([reference](https://arxiv.org/abs/1611.05431)).
+Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
+
+!!! warning
+    `ResNeXt` does not currently support pretrained weights.
+
+See also [`Metalhead.resnext`](#).
+"""
+function ResNeXt(; config::Int = 50, cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
```
darsnack, line 111: `config` should be a positional argument, and `cardinality` and `width` should be part of the configuration dictionary.

test/runtests.jl
```diff
@@ -50,13 +50,28 @@ end
 
   @testset "Shortcut C" begin
     m = Metalhead.resnet(Metalhead.basicblock, :C;
-                         channel_config = [1, 1],
-                         block_config = [2, 2, 2, 2])
+      channel_config = [1, 1],
+      block_config = [2, 2, 2, 2])
```
darsnack, line 54: Try not to change the other code.

test/runtests.jl
```diff
@@ -50,13 +50,28 @@ end
...
 
     @test size(m(rand(Float32, 256, 256, 3, 2))) == (1000, 2)
   end
 end
 
+@testset "ResNeXt" begin
+  @testset for depth in [50, 101]
+    model = ResNeXt
+    m = model(config = depth)
```
darsnack, line 63:
```suggestion
    m = ResNeXt(config = depth)
```

darsnack (changes requested):

darsnack:
Looking good, just a few more things to clear up.

src/resnext.jl
theabhirath: That doesn't really match the image though. In that case, the inner channels should always be 128, whereas the image explicitly shows the inner channels increasing for every block. And plus that configuration does not match the parameters specified in the paper, only this one does. The params and FLOPs line up with this specific implementation. But let me know if I'm missing something

test/runtests.jl
theabhirath: Whoops, must've snuck in there accidentally, apologies

darsnack: No problem!

test/runtests.jl
theabhirath: I kept the original style because both `model` and `m` are being used for tests...should I just do away with the model definition and replace it everywhere by `ResNeXt`?

src/resnext.jl
theabhirath: This one is tricky, because what the original ResNeXt paper allows for is changes to cardinality and width even while maintaining a single configuration of layers - they've given versions that are 64x4d, for instance. I wanted to allow for that. I'll change config to be a positional argument, though.

src/resnext.jl
darsnack: Oh you're right, I was completely focused on the figures, but the table does indeed have an increasing inner dimension. What I would suggest then is still keeping the `inner = cardinality * width` approach since that matches how the bottleneck is described in the paper. Then in main `resnext` function, we can widen the `width` passed in by `widen_factor` just like we do for `baseplanes`.

src/resnext.jl
```diff
@@ -0,0 +1,117 @@
...
+        # downsample the residual path if this is the first repetition of a block
+        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, i != 1),
+            skip_projection(inplanes, outplanes, i != 1)))
+        # push remaining skip connections on using second residual
+        inplanes = outplanes
+        for _ in 2:nrepeats
+            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, widen_factor, false),
+                skip_identity(inplanes, outplanes, false)))
+        end
+        baseplanes = outplanes
```
darsnack, line 60:
```suggestion
        baseplanes = outplanes
        width *= widen_factor
```

darsnack (changes requested):

test/runtests.jl
darsnack: Yeah just for ResNeXt tests. The old style was only necessary cause the function names were different for each configuration. In the new approach, we know it will always be `ResNeXt` with the depth argument varying.

theabhirath:
Should I also shift the ResNet utility functions `skip_identity` and `skip_projection` over to `utilities.jl`? I've used them in ResNeXt and a couple of other models that I'm drafting without it being clear that they're actually in the ResNet file

darsnack:
Yes, that makes sense

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+"""
+function resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
+    stride = downsample ? 2 : 1
+
+    hidden_channels = cardinality * width
+
+    return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
+        conv_bn((3, 3), hidden_channels, hidden_channels;
+            stride = stride, pad = 1, bias = false, groups = cardinality)...,
+        conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
```
darsnack, line 22:
```suggestion
    return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
                 conv_bn((3, 3), hidden_channels, hidden_channels;
                         stride = stride, pad = 1, bias = false, groups = cardinality)...,
                 conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
+"""
+    resnextblock(inplanes, outplanes, cardinality, downsample = false)
```
darsnack, line 2:
```suggestion
    resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+
+# Arguments
+- `cardinality`: the number of groups to use for the convolution
+- `width`: the number of feature maps in each group in the bottleneck
+- `connection`: the binary function applied to the output of residual and skip paths in a block
+- `block_config`: a list of the number of residual blocks at each stage
+- `nclasses`: the number of output classes
+"""
+function resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @. relu(x) + relu(y);
+                block_config, nclasses = 1000)
```
darsnack, line 40:
```suggestion
                 block_config, nclasses = 1000)
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
+        conv_bn((3, 3), hidden_channels, hidden_channels;
+            stride = stride, pad = 1, bias = false, groups = cardinality)...,
+        conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
+end
+
+"""
+    resnext(connection = (x, y) -> @. relu(x) + relu(y); channel_config, block_config, 
+            cardinality, nclasses = 1000)
```
darsnack, line 27:
```suggestion
    resnext(cardinality, width, widen_factor, connection = (x, y) -> @. relu(x) + relu(y);
            block_config, nclasses = 1000)
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+    
+Create a ResNeXt model
+([reference](https://arxiv.org/abs/1611.05431)).
+
+# Arguments
+- `cardinality`: the number of groups to use for the convolution
+- `width`: the number of feature maps in each group in the bottleneck
```
darsnack, line 34: Add `widen_factor`

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+    layers = []
+    append!(layers, conv_bn((7, 7), 3, inplanes; stride = 2, pad = (3, 3)))
+    push!(layers, MaxPool((3, 3), stride = (2, 2), pad = (1, 1)))
+    for (i, nrepeats) in enumerate(block_config)
+        # output planes within a block
+        outplanes = baseplanes * widen_factor
+        # push first skip connection on using first residual
+        # downsample the residual path if this is the first repetition of a block
+        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
+            skip_projection(inplanes, outplanes, i != 1)))
```
darsnack, line 52:
```suggestion
        push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, i != 1),
                                           skip_projection(inplanes, outplanes, i != 1)))
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+        # push remaining skip connections on using second residual
+        inplanes = outplanes
+        for _ in 2:nrepeats
+            push!(layers, Parallel(connection, resnextblock(inplanes, outplanes, cardinality, width, false),
+                skip_identity(inplanes, outplanes, false)))
```
darsnack, line 57: Same as above

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+        end
+        baseplanes = outplanes
+        width *= widen_factor
+    end
+
+    return Chain(Chain(layers...),
+        Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
```
darsnack, line 64:
```suggestion
    return Chain(Chain(layers...),
                 Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
```

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+
+"""
+    ResNeXt(cardinality, width; block_config, nclasses = 1000)
+    
+Create a ResNeXt model
+([reference](https://arxiv.org/abs/1611.05431)).
+
+# Arguments
+- `cardinality`: the number of groups to use for the convolution
+- `width`: the number of feature maps in each group in the bottleneck
```
darsnack, line 75: Add `block_config`

src/resnext.jl
```diff
@@ -0,0 +1,118 @@
...
+const resnext_config = Dict(
+    50 => (3, 4, 6, 3),
+    101 => (3, 4, 23, 3),
+    152 => (3, 8, 36, 3)
+)
+
+"""
+    ResNeXt(config::Int = 50; cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
+
+Create a ResNeXt model with specified configuration. Currently supported values are (50, 101).
```
darsnack, line 103:
```suggestion
Create a ResNeXt model with specified configuration. Currently supported values for `config` are (50, 101).
```

test/runtests.jl
```diff
@@ -50,13 +50,27 @@ end
 
   @testset "Shortcut C" begin
     m = Metalhead.resnet(Metalhead.basicblock, :C;
-                         channel_config = [1, 1],
-                         block_config = [2, 2, 2, 2])
+                        channel_config = [1, 1],
+                        block_config = [2, 2, 2, 2])
```
darsnack, line 54:
```suggestion
                         channel_config = [1, 1],
                         block_config = [2, 2, 2, 2])
```

darsnack (changes requested):
Okay looks great, mostly minor formatting changes at this point and fixing docstrings. One annoying change is to use an indent of 2 spaces like the rest of the repo.

theabhirath:
I _think_ that's everything....

src/resnext.jl
```diff
@@ -0,0 +1,121 @@
+"""
+  resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
```
darsnack, line 2:
```suggestion
    resnextblock(inplanes, outplanes, cardinality, width, downsample = false)
```

src/resnext.jl
```diff
@@ -0,0 +1,121 @@
...
+  hidden_channels = cardinality * width
+
+  return Chain(conv_bn((1, 1), inplanes, hidden_channels; stride = 1, bias = false)...,
+               conv_bn((3, 3), hidden_channels, hidden_channels;
+                        stride = stride, pad = 1, bias = false, groups = cardinality)...,
+               conv_bn((1, 1), hidden_channels, outplanes; stride = 1, bias = false)...)
+end
+
+"""
+  resnext(cardinality, width, widen_factor, connection = (x, y) -> @. relu(x) + relu(y);
```
darsnack, line 26:
```suggestion
    resnext(cardinality, width, widen_factor = 2, connection = (x, y) -> @. relu(x) + relu(y);
```

src/resnext.jl
```diff
@@ -0,0 +1,121 @@
...
+    # double width after every cluster of blocks
+    width *= widen_factor
+  end
+
+  return Chain(Chain(layers...),
+               Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(inplanes, nclasses)))
+end
+
+"""
+  ResNeXt(cardinality, width; block_config, nclasses = 1000)
```
darsnack, line 70:
```suggestion
    ResNeXt(cardinality, width; block_config, nclasses = 1000)
```

src/resnext.jl
```diff
@@ -0,0 +1,121 @@
...
+classifier(m::ResNeXt) = m.layers[2]
+
+const resnext_config = Dict(
+  50 => (3, 4, 6, 3),
+  101 => (3, 4, 23, 3),
+  152 => (3, 8, 36, 3)
+)
+
+"""
+  ResNeXt(config::Int = 50; cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
```
darsnack, line 104:
```suggestion
    ResNeXt(config::Int = 50; cardinality = 32, width = 4, pretrain = false, nclasses = 1000)
```

darsnack (changes requested):
Almost... 😅

darsnack (approved):
Looks great! Will merge once CI is green. Thank you @theabhirath for seeing this through! Looking forward to more contributions.

theabhirath:
> Looks great! Will merge once CI is green. Thank you @theabhirath for seeing this through! Looking forward to more contributions.

Thanks for walking me through! I'm gonna be trying for some other missing models like MobileNet and EfficientNet next, I'll need the help 😅.

> Almost... 😅. FluxML's style is needlessly frustrating to follow.

By any chance, is there any auto-formatter? I already changed VSCode to have indent as two spaces for the future, but for the other style guidelines?

darsnack:
> Thanks for walking me through! I'm gonna be trying for some other missing models like MobileNet and EfficientNet next, I'll need the help 😅.

Looking forward to it!

> By any chance, is there any auto-formatter? I already changed VSCode to have indent as two spaces for the future, but for the other style guidelines?

Unfortunately, no. But the only other main one I noticed was when you break a long list of arguments onto a new line. The next line should line up with the first argument in the previous line (not just a 1 indent like other languages).

DhairyaLGandhi:
Great job @theabhirath ! Also see https://github.com/pxl-th/EfficientNet.jl and https://github.com/pxl-th/Segmentation.jl/issues/1 for EfficientNet. Would love to see these in Metalhead!

PR #101, opened 2022-01-18
Implementation of MobileNetv2 and MobileNetv3

theabhirath:
This is an implementation of the two most recent and widely-used MobileNet models, [MobileNetv2](https://arxiv.org/abs/1801.04381) and [MobileNetv3](https://arxiv.org/pdf/1905.02244).


lorenzoh:
Very nice! Have you tested the models on GPU? I recall trying to implement these a way back and the grouped convolutions not working yet.

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+      inplanes = outplanes
+    end
+  end
+
+  # building last several layers
+  outplanes = (width_mult > 1.0) ? _make_divisible(1280 * width_mult, width_mult == 0.1 ? 4 : 8) : 1280
+
+  return Chain(Chain(layers...),
+               conv_bn((1, 1), inplanes, outplanes, relu6, bias = false)...,
+               AdaptiveMeanPool((1, 1)), flatten, Dense(outplanes, nclasses))
```
darsnack, line 91:
```suggestion
  return Chain(Chain(layers...,
                     conv_bn((1, 1), inplanes, outplanes, relu6, bias = false)...),
               Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(outplanes, nclasses)))
```

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+  configs = [
+    # t, c, n, s
+    [1, 16, 1, 1],
+    [6, 24, 2, 2],
+    [6, 32, 3, 2],
+    [6, 64, 4, 2],
+    [6, 96, 3, 1],
+    [6, 160, 3, 2],
+    [6, 320, 1, 1],
+  ]
```
darsnack, line 70: Even though MobileNetV2 is a restricted architecture, should we allow this config to be passed in?

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+  for (t, c, n, s) in configs
+    outplanes = _make_divisible(c * width_mult, width_mult == 0.1 ? 4 : 8)
+    for i in 1:n
+      push!(layers, invertedresidualv2(inplanes, outplanes, i == 1 ? s : 1, t))
+      inplanes = outplanes
+    end
+  end
+
+  # building last several layers
+  outplanes = (width_mult > 1.0) ? _make_divisible(1280 * width_mult, width_mult == 0.1 ? 4 : 8) : 1280
```
darsnack, line 87: Similarly, should we let the 1280 be customizable?

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+backbone(m::MobileNetv2) = m.layers[1]
+classifier(m::MobileNetv2) = m.layers[2:end]
+
+# MobileNetv3
+
+"""
+Hard-Swish activation function
+([reference](https://arxiv.org/abs/1905.02244)).
+"""
+@inline h_swish(x) = x * hardσ(x)
```
darsnack, line 132: Makes more sense to make a PR to NNlib.jl with this.

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+
+# Arguments
+- `configs`: a list of configurations for each layer that details 
+- `mode`: 'large' or 'small'
+- `width_mult`: Controls the number of channels in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `nclasses`: the number of output classes
+"""
+function mobilenetv3(configs, mode, width_mult; nclasses = 1000)
+  @assert mode in ["large", "small"] "`mode` has to be either \"large\" or \"small\""
```
darsnack, line 201: Prefer to use `Symbol`s instead of `String`s.

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+    # inverted residual layers
+    outplanes = _make_divisible(c * width_mult, 8)
+    global explanes = _make_divisible(inplanes * t, 8)
+    push!(layers, invertedresidualv3(inplanes, explanes, outplanes, k, s, use_se, use_hs))
+    inplanes = outplanes
+  end
+
+  # building last several layers
+  output_channel = Dict("large" => 1280, "small" => 1024)
+  output_channel = width_mult > 1.0 ? _make_divisible(output_channel[mode] * width_mult, 8) : output_channel[mode]
```
darsnack, line 225:
```suggestion
  output_channel = (mode == "large") ? 1280 : 1024
  output_channel = width_mult > 1.0 ? _make_divisible(output_channel * width_mult, 8) : output_channel
```

darsnack, line 225: Similar comment here too about making the output channels customizable, then restricting it at a higher level in the API.

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+  classifier = Chain(
+    Dense(explanes, output_channel, h_swish),
+    Dropout(0.2),
+    Dense(output_channel, nclasses),
+  )
+
+  return Chain(Chain(layers...),
+               conv_bn((1, 1), inplanes, explanes, h_swish, bias = false)...,
+               AdaptiveMeanPool((1, 1)), flatten, classifier)
```
darsnack, line 234:
```suggestion
  classifier = (
    Dense(explanes, output_channel, h_swish),
    Dropout(0.2),
    Dense(output_channel, nclasses),
  )

  return Chain(Chain(layers...,
                     conv_bn((1, 1), inplanes, explanes, h_swish, bias = false)...),
               Chain(AdaptiveMeanPool((1, 1)), flatten, classifier...))
```

test/runtests.jl
```diff
@@ -105,3 +105,32 @@ end
...
+      @test size(m(rand(Float32, 224, 224, 3, 2))) == (1000, 2)
+      if MobileNetv3 in PRETRAINED_MODELS
+        @test (MobileNetv3(; mode, pretrain = true); true)
+      else
+        @test_throws ArgumentError MobileNetv3(; mode, pretrain = true)
+      end
+      @test_skip gradtest(m, rand(Float32, 224, 224, 3, 2))
+    end
+  end
+end
```
darsnack, line 136:
```suggestion
end

```

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+  for (k, t, c, use_se, use_hs, s) in configs
+    # converting the stored parameters to their respective types
+    k = convert(Int, k)
+    c = convert(Int, c)
+    use_se = convert(Bool, use_se)
+    use_hs = convert(Bool, use_hs)
+    s = convert(Int, s)
+    # inverted residual layers
+    outplanes = _make_divisible(c * width_mult, 8)
+    global explanes = _make_divisible(inplanes * t, 8)
```
darsnack, line 218: Is this so that you can access `explanes` outside the loop? Instead of using `global`, you should initialize it before the loop (e.g. `explanes = 0`). Or even better, leave it as is without `global`, then outside the loop use `Flux.outputsize` to get the size for use later in the code.

darsnack (changes requested):
Looks pretty good on the first pass.

In general, for the lowest level function (e.g. `mobilenetv2`), we want to cast as wide a net as possible when it comes to configuration. If you find that it is possible to loop over a configuration dictionary, then it should probably be a user-specified parameter. The higher level functions like `MobileNetV2` exist to provide convenience constructors for passing in the correct configs.

src/mobilenet.jl
```diff
@@ -0,0 +1,304 @@
...
+  append!(layers, conv_bn((3, 3), 3, inplanes, h_swish; stride = 2))
+
+  # building inverted residual blocks
+  for (k, t, c, use_se, use_hs, s) in configs
+    # converting the stored parameters to their respective types
+    k = convert(Int, k)
+    c = convert(Int, c)
+    use_se = convert(Bool, use_se)
+    use_hs = convert(Bool, use_hs)
+    s = convert(Int, s)
```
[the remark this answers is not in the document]
theabhirath: Whoops that was a bit of hacky code I used because I copied the `configs` dict over and didn't want to spend time fixing the types 😅. I'll fix that up

src/mobilenet.jl
theabhirath: Yeah I missed `Flux.outputsize`, that's an amazing function xD. Will make the change

theabhirath:
> Looks pretty good on the first pass.
> 
> In general, for the lowest level function (e.g. `mobilenetv2`), we want to cast as wide a net as possible when it comes to configuration. If you find that it is possible to loop over a configuration dictionary, then it should probably be a user-specified parameter. The higher level functions like `MobileNetV2` exist to provide convenience constructors for passing in the correct configs.

Okay I'll fix the formatting errors (tried my best, some seem to have gotten past again). For the configurations, I initially thought of passing the `configs` in but I ultimately decided not to because:

1. It's a fairly large configuration and I'm not quite sure how we'd deal with all the edge cases for the possible values of the kernel size etc. and 
2. Almost every use case I found changed the width multiplier, not the configs because that's a more fundamental value to the network. 

Lemme know if you still think I should make that change though, I'll go ahead.

src/mobilenet.jl
theabhirath: Have done so, will use that version once it gets merged

src/mobilenet.jl
theabhirath: Have done. I'm a little new to the concept of metaprogramming though, so what exactly is the difference here? And why are `Symbols` preferred?

src/mobilenet.jl
theabhirath: I've changed the API to allow for this and added a note in the documentation, hopefully that should prevent people from causing errors with the configurations

src/mobilenet.jl
theabhirath: https://github.com/FluxML/NNlib.jl/pull/376 takes care of this. But given that it's in `master` and not in the current release, keeping this line of code for now.

src/mobilenet.jl
theabhirath: Fixed this in the latest API

src/mobilenet.jl
theabhirath: Fixed

theabhirath:
> > Looks pretty good on the first pass.
> > In general, for the lowest level function (e.g. `mobilenetv2`), we want to cast as wide a net as possible when it comes to configuration. If you find that it is possible to loop over a configuration dictionary, then it should probably be a user-specified parameter. The higher level functions like `MobileNetV2` exist to provide convenience constructors for passing in the correct configs.
> 
> Okay I'll fix the formatting errors (tried my best, some seem to have gotten past again). For the configurations, I initially thought of passing the `configs` in but I ultimately decided not to because:
> 
> 1. It's a fairly large configuration and I'm not quite sure how we'd deal with all the edge cases for the possible values of the kernel size etc. and
> 2. Almost every use case I found changed the width multiplier, not the configs because that's a more fundamental value to the network.
> 
> Lemme know if you still think I should make that change though, I'll go ahead.

Decided to just document it well and hope that the users don't try and do something too dangerous with the models 😂. There's examples in the code though so hopefully it isn't very difficult to follow

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+"""
+    mobilenetv2(width_mult, configs; max_width = 1280, nclasses = 1000)
+
+Create a MobileNetv2 model.
+([reference](https://arxiv.org/abs/1801.04381)).
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `configs`: A "list of lists" configuration for each layer that details:
```
darsnack, line 52:
```suggestion
- `configs`: A "vector of vectors" configuration for each layer that details:
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+
+"""
+    mobilenetv2(width_mult, configs; max_width = 1280, nclasses = 1000)
+
+Create a MobileNetv2 model.
+([reference](https://arxiv.org/abs/1801.04381)).
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
```
darsnack, line 51:
```suggestion
- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+([reference](https://arxiv.org/abs/1801.04381)).
+Set `pretrain` to `true` to load the pretrained weights for ImageNet. 
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `pretrain`: Whether to load the pre-trained weights for ImageNet
+- `nclasses`: The number of output classes
+"""
+function MobileNetv2(width_mult::Float64 = 1.0; pretrain = false, nclasses = 1000)
```
darsnack, line 114: Does this have to be a `Float64`?

```suggestion
function MobileNetv2(width_mult = 1.0; pretrain = false, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+
+"""
+Hard-Swish activation function
+([reference](https://arxiv.org/abs/1905.02244)).
+"""
+@inline h_swish(x) = x * hardσ(x)
+
+"""
+Squeeze and Excitation layer used by MobileNetv3
+([reference](https://arxiv.org/abs/1905.02244)).
```
darsnack, line 138:
```suggestion
    selayer(channels, reduction = 4)

Squeeze and Excitation layer used by MobileNetv3
([reference](https://arxiv.org/abs/1905.02244)).
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+  # building last several layers
+  outplanes = (width_mult > 1.0) ? _make_divisible(max_width * width_mult, width_mult == 0.1 ? 4 : 8) : max_width
+
+  return Chain(Chain(layers...,
+                     conv_bn((1, 1), inplanes, outplanes, relu6, bias = false)...),
+               Chain(AdaptiveMeanPool((1, 1)), flatten, Dense(outplanes, nclasses)))
+end
+
+# Layer configurations for MobileNetv2
+mobilenetv2_configs = [
```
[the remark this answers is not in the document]
darsnack: Could make this a vector of tuples for consistency.

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+  )
+
+  return Chain(Chain(layers...,
+                     conv_bn((1, 1), inplanes, explanes, h_swish, bias = false)...),
+               Chain(AdaptiveMeanPool((1, 1)), flatten, classifier...))
+end
+
+# Configurations for small and large mode for MobileNetv3
+mobilenetv3_configs = Dict(
+  "small" => [
```
darsnack, line 233:
```suggestion
  :small => [
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+    (5, 4, 40, true, true, 2),
+    (5, 6, 40, true, true, 1),
+    (5, 6, 40, true, true, 1),
+    (5, 3, 48, true, true, 1),
+    (5, 3, 48, true, true, 1),
+    (5, 6, 96, true, true, 2),
+    (5, 6, 96, true, true, 1),
+    (5, 6, 96, true, true, 1),
+  ], 
+  "large" => [
```
darsnack, line 247:
```suggestion
  :large => [
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+  - `use_hs::Bool` - Whether to use Hard-Swish activation function
+  - `s::Int` - The stride of the convolutional kernel
+- `mode`: Can be set to be either 'large' or 'small' to decide the size of the model
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `max_width`: The maximum number of feature maps in any layer of the network
+- `nclasses`: the number of output classes
+"""
+function mobilenetv3(configs, mode, width_mult; max_width = 1024, nclasses = 1000)
+  @assert Symbol(mode) in [:large, :small] "`mode` has to be either \"large\" or \"small\""
```
darsnack, line 201: It doesn't look like `mode` is used anywhere in the function anymore?

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+    MobileNetv3(width_mult::Float64 = 1.0; mode = small; pretrain = false, nclasses = 1000)
+
+Create a MobileNetv3 model with the specified configuration.
+([reference](https://arxiv.org/abs/1905.02244)).
+Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `mode`: 'small' or 'large' for the size of the model
```
darsnack, line 282:
```suggestion
- `mode`: :small or :large for the size of the model (see paper).
- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper).
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+  ]
+)
+
+# Model definition for MobileNetv3
+struct MobileNetv3
+  layers
+end
+
+"""
+    MobileNetv3(width_mult::Float64 = 1.0; mode = small; pretrain = false, nclasses = 1000)
```
darsnack, line 273:
```suggestion
    MobileNetv3(mode = :small, width_mult = 1.0; pretrain = false, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `mode`: 'small' or 'large' for the size of the model
+- `pretrain`: whether to load the pre-trained weights for ImageNet
+- `nclasses`: the number of output classes
+"""
+function MobileNetv3(width_mult::Float64 = 1.0; mode = "small", pretrain = false, nclasses = 1000)
```
darsnack, line 286:
```suggestion
function MobileNetv3(mode = :small, width_mult = 1.0; pretrain = false, nclasses = 1000)
```

src/mobilenet.jl
theabhirath: Thank you so much!

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+  @assert Symbol(mode) in [:large, :small] "`mode` has to be either \"large\" or \"small\""
```
darsnack, line 287:
```suggestion
  @assert mode in [:large, :small] "`mode` has to be either :large or :small"
```

src/mobilenet.jl
```diff
@@ -0,0 +1,300 @@
...
+
+  max_width = (mode == "large") ? 1280 : 1024
```
darsnack, line 289:
```suggestion
  max_width = (mode == :large) ? 1280 : 1024
```

test/runtests.jl
```diff
@@ -105,3 +105,32 @@ end
...
+      @test (MobileNetv2(pretrain = true); true)
+    else
+      @test_throws ArgumentError MobileNetv2(pretrain = true)
+    end
+    @test_skip gradtest(m, rand(Float32, 224, 224, 3, 2))
+  end
+
+  @testset "MobileNetv3" verbose = true begin
+    @testset for mode in ["small", "large"]
+      m = MobileNetv3(; mode)
```
darsnack, line 125:
```suggestion
      m = MobileNetv3(mode)
```

test/runtests.jl
```diff
@@ -105,3 +105,32 @@ end
...
+
+      @test size(m(rand(Float32, 224, 224, 3, 2))) == (1000, 2)
+      if MobileNetv3 in PRETRAINED_MODELS
+        @test (MobileNetv3(; mode, pretrain = true); true)
+      else
+        @test_throws ArgumentError MobileNetv3(; mode, pretrain = true)
```
darsnack, line 131:
```suggestion
        @test_throws ArgumentError MobileNetv3(mode; pretrain = true)
```

darsnack (changes requested):
Okay, shaping up pretty well. I agree, let's not include the config as part of the arguments for MobileNetV2 for now.

src/mobilenet.jl
darsnack: I don't really think there's a big difference here, but Julia conventions typically use `Symbol`s for arguments of this form. `Symbol`s do have a special meaning in metaprogramming, but in this context, they are effectively just another data type (i.e. we are not doing any metaprogramming).

src/mobilenet.jl
theabhirath: Nope, realised that the function call works just fine without. Fixed

src/mobilenet.jl
theabhirath: Yep, got eliminated amongst the shuffle of the configs

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `pretrain`: Whether to load the pre-trained weights for ImageNet
+- `nclasses`: The number of output classes
+
+See also [`Metalhead.mobilenetv2`](#).
+"""
+function MobileNetv2(width_mult::Float64 = 1.0; pretrain = false, nclasses = 1000)
```
darsnack, line 115:
```suggestion
function MobileNetv2(width_mult = 1; pretrain = false, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+  - `t::Int` - The number of output feature maps for a given block
+  - `use_se::Bool` - Whether to use Squeeze and Excitation layer
+  - `use_hs::Bool` - Whether to use Hard-Swish activation function
+  - `s::Int` - The stride of the convolutional kernel
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
+- `max_width`: The maximum number of feature maps in any layer of the network
+- `nclasses`: the number of output classes
+"""
+function mobilenetv3(configs, width_mult; max_width = 1024, nclasses = 1000)
```
darsnack, line 196: I see the reasoning for `config` first in this case, but I think consistency with v2 outweighs it.
```suggestion
function mobilenetv3(width_mult, configs; max_width = 1024, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+# Arguments
+- `mode`: :small or :large for the size of the model (see paper).
+- `width_mult`: Controls the number of output feature maps in each block
+                (with 1.0 being the default in the paper).
+- `pretrain`: whether to load the pre-trained weights for ImageNet
+- `nclasses`: the number of output classes
+
+See also [`Metalhead.mobilenetv3`](#).
+"""
+function MobileNetv3(mode::Symbol = :small, width_mult = 1.0; pretrain = false, nclasses = 1000)
```
darsnack, line 282:
```suggestion
function MobileNetv3(mode = :small, width_mult = 1; pretrain = false, nclasses = 1000)
```

darsnack (changes requested):
Few small changes. Otherwise this is good to go.

src/mobilenet.jl
theabhirath: I kept it as 1.0 so that it's clear to the user that `width_mult` is in general a floating point value. Will change it if you think that's apparent, though

src/mobilenet.jl
theabhirath: This one runs into problems when I remove the ::Symbol. It doesn't call the function and just displays as `MobileNetv3(:small)`. Hence the type annotation. Same comment for the `width_mult` as above

src/mobilenet.jl
darsnack: Sure though we can do that in the docstring and not the code (that's the part the user will see).

src/mobilenet.jl
darsnack: Oh right cause it conflicts with the default constructor for the struct. We can leave them in then, but widen the type of `widen_mult` to `Number` instead of `Float64`.

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+                            bias = false, stride, pad = (kernel_size - 1) ÷ 2, groups = hidden_planes)...,
+                            use_se ? selayer(hidden_planes) : identity,
+                   conv_bn((1, 1), hidden_planes, outplanes, identity; bias = false)...)
+  end
+
+  (stride == 1 && inplanes == outplanes) ? SkipConnection(invres, +) : invres
+end
+
+"""
+    mobilenetv3(configs, width_mult; max_width = 1024, nclasses = 1000)
```
darsnack, line 178:
```suggestion
    mobilenetv3(width_mult, configs; max_width = 1024, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+# Arguments
+- `configs`: a "list of tuples" configuration for each layer that details:
+  - `k::Int` - The size of the convolutional kernel
+  - `c::Float` - The multiplier factor for deciding the number of feature maps in the hidden layer
+  - `t::Int` - The number of output feature maps for a given block
+  - `use_se::Bool` - Whether to use Squeeze and Excitation layer
+  - `use_hs::Bool` - Whether to use Hard-Swish activation function
+  - `s::Int` - The stride of the convolutional kernel
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper.
```
darsnack, line 192:
```suggestion

- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper;
                 this is usually a value between 0.1 and 1.4)
- `configs`: a "list of tuples" configuration for each layer that details:
  - `k::Int` - The size of the convolutional kernel
  - `c::Float` - The multiplier factor for deciding the number of feature maps in the hidden layer
  - `t::Int` - The number of output feature maps for a given block
  - `use_se::Bool` - Whether to use Squeeze and Excitation layer
  - `use_hs::Bool` - Whether to use Hard-Swish activation function
  - `s::Int` - The stride of the convolutional kernel
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+  (6, 320, 1, 1)
+]
+
+# Model definition for MobileNetv2
+struct MobileNetv2
+  layers
+end
+
+"""
+    MobileNetv2(width_mult::Number = 1; pretrain = false, nclasses = 1000)
```
darsnack, line 101:
```suggestion
    MobileNetv2(width_mult = 1.0; pretrain = false, nclasses = 1000)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+
+Create a MobileNetv2 model with the specified configuration.
+([reference](https://arxiv.org/abs/1801.04381)).
+Set `pretrain` to `true` to load the pretrained weights for ImageNet. 
+
+# Arguments
+- `width_mult`: Controls the number of feature maps in each layer, with 1.0 being the original
+  model as detailed in the paper. This is usually a floating point value in between 0.1 and 1.4.
```
darsnack, line 109:
```suggestion
- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper;
                 this is usually a value between 0.1 and 1.4)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+
+"""
+    mobilenetv2(width_mult, configs; max_width = 1280, nclasses = 1000)
+
+Create a MobileNetv2 model.
+([reference](https://arxiv.org/abs/1801.04381)).
+
+# Arguments
+- `width_mult`: Controls the number of output feature maps in each block
+                (with 1.0 being the default in the paper)
```
darsnack, line 51:
```suggestion

- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper;
                 this is usually a value between 0.1 and 1.4)
```

src/mobilenet.jl
```diff
@@ -0,0 +1,296 @@
...
+    MobileNetv3(mode::Symbol = :small, width_mult::Number = 1; pretrain = false, nclasses = 1000)
+
+Create a MobileNetv3 model with the specified configuration.
+([reference](https://arxiv.org/abs/1905.02244)).
+Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
+
+# Arguments
+- `mode`: :small or :large for the size of the model (see paper).
+- `width_mult`: Controls the number of output feature maps in each block (with 1.0 being the 
+  default in the paper). This is usually a floating point value in between 0.1 and 1.4.
```
darsnack, line 276:
```suggestion
- `width_mult`: Controls the number of output feature maps in each block
                (with 1.0 being the default in the paper;
                 this is usually a value between 0.1 and 1.4)
```

darsnack (changes requested):

darsnack (approved):
Looks good, will merge once CI is green! Thanks for another great PR @theabhirath.

darsnack:
@theabhirath some of the old changes were reverted by the force push

theabhirath:
Shoot, must've missed those. The tests were failing because we changed the API for the MobileNetv3 so I had to go back and fix em. This seems fine now though?

darsnack:
Okay just waiting on checks now

theabhirath:
I'm a little confused how this happened, honestly, given that I only squashed my commits, so the earlier changes should've still been preserved

darsnack:
Yeah not sure. Just minor docstring changes anyways.

PR #103, opened 2022-01-29
Implementation of MLPMixer

theabhirath:
This is an implementation of one of the many models in the wake of the ViT explosion, [MLPMixer](https://arxiv.org/pdf/2105.01601).

There's two things I wanted to clarify:
1. I added TensorCast as a dep because einops like operations are very commonplace in ViT model implementations and I thought it would be easier to work with that. If you think that I should change that up somehow or maybe just use standard array operations, lemme know and I'll revert - it's very painful though, I tried already 🥲
2. I _also_ made some changes to the organisation of the repo (might be a bit of an understatement) because I thought it would be better to separate out the CNN models and the ViT models (I have some others in the works) but lemme know if you think it's overkill and I'll revert to the original repo structure.


darsnack:
Re-organizing the repo into CNNs and ViTs makes sense, but I would avoid introducing submodules. I would just have two folders and put all the includes in `Metalhead.jl` (grouped by type to make it look nice).

I like TensorCast.jl, but I want to avoid taking on a dep if we can. I'll have to go through and see if we can rewrite using standards ops without being too messy.

Now, we got ViTs, exciting!

darsnack:
I forgot to ask this on the last PR, but can you update the table in the README with the models from the previous PRs + this one?

CarloLucibello:
I don't think MLPMixer can be classified as a visual transformer, but maybe we could argue that this separation is fine since the spirit is the same

theabhirath:
> Re-organizing the repo into CNNs and ViTs makes sense, but I would avoid introducing submodules. I would just have two folders and put all the includes in `Metalhead.jl` (grouped by type to make it look nice).
> 
Will do.
> I like TensorCast.jl, but I want to avoid taking on a dep if we can. I'll have to go through and see if we can rewrite using standards ops without being too messy.
> 
Yeah the MLPMixer one isn't that bad, it's just two places, but for the standard ViT and for other versions like LeViT and Swin it gets progressively worse.

> I forgot to ask this on the last PR, but can you update the table in the README with the models from the previous PRs + this one?

Will do, I wasn't sure whether or not to do it because it wasn't tagged as a release yet xD.

> I don't think MLPMixer can be classified as a visual transformer, but maybe we could argue that this separation is fine since the spirit is the same

Yeah they share quite a few things with the patches and embeddings design, and the original [JAX repository](https://github.com/google-research/vision_transformer) released the code together so I think it's better to keep them together

darsnack:
> Will do, I wasn't sure whether or not to do it because it wasn't tagged as a release yet xD.

Only the main branch README will change. The docs will still reflect the tagged version.

> I don't think MLPMixer can be classified as a visual transformer, but maybe we could argue that this separation is fine since the spirit is the same

True, but it doesn't really fit any category. Once the submodules are removed, this distinction will only affect our code organization and not the user interfaces.

theabhirath:
That's odd, the CI seems to be failing only on Linux and I don't see why...the error message says the file doesn't exist but it very much does 🤨

theabhirath:
Nope, still crashing and burning. I can't see what's wrong, I'm on a Mac but the file paths seem okay when I execute it locally?

theabhirath:
Yeah some sort of weird renaming issue - I'd done it on the local but it didn't seem to have been pushed to the remote somehow. Fixed now

darsnack:
Can you rebase with the latest changes? Shouldn't affect your work here.

ToucheSir:
Bikeshedding the folder structure, I don't think there needs to be a premature assignment of MLP Mixer into a specific category (e.g. where would DeiT fit in under this scheme?). Keeping it at the top level or in an "other" directory would be fine. If/when we get attention-based models, those can get their own dir.

theabhirath:
> Bikeshedding the folder structure, I don't think there needs to be a premature assignment of MLP Mixer into a specific category (e.g. where would DeiT fit in under this scheme?). Keeping it at the top level or in an "other" directory would be fine. If/when we get attention-based models, those can get their own dir.

I've make the change for MLPMixer to be put in an "other" directory, then. I think DeiT would slot into the ViT folder given that the paper explicitly refers to it as such ([FAIR's repo](https://github.com/facebookresearch/deit) released all their ViT-based models together too) but I get the reason for the MLPMixer contention so I've taken care of that

theabhirath:
Playing around with this model, I realised that the `show` functionality doesn't work as expected because I've defined a custom model instead of writing the layers with say `Chain` or `SkipConnection`, which are already present in Flux and thus likely implement `show` on their own. Is there an easier way to get it done than having to write it layer by layer for the custom operations? There's also the use of TensorCast, which I imagine will complicate things slightly

darsnack:
I have some suggestions which should resolve that issue. I've just been busy over the last few days, but I'll review both new PRs later this evening.

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,87 @@
...
+                                  channel_mix)),) for _ in 1:depth]
+
+  MLPMixer(channels,
+           planes,
+           patch_size,
+           num_patches,
+           token_mix,
+           channel_mix,
+           layers,
+           nclasses)
```
DhairyaLGandhi, line 75: Do we need to store a bunch of this? I don't think it is used in the forwards pass.

src/vit-like/mlpmixer.jl
```diff
@@ -0,0 +1,87 @@
...
+           layers,
+           nclasses)
+end
+
+function (m::MLPMixer)(x)
+  p = m.patch_size
+  @cast x[(h2, w2, c), (h, w), b] := x[(h, h2), (w, w2), c, b]  h2 in 1:p, w2 in 1:p
+  x = Dense((m.patch_size ^ 2) * m.channels, m.planes)(x)
+  x = Chain(LayerNorm(m.planes), m.layers...)(x)
+  @reduce x[b, c] := mean(n) x[b, n, c]
```
darsnack: The global average pooling can simply be done with `mean(x; dims = ...)`. Though I'm confused how the batch dimension ended up first in the current code.

darsnack (changes requested):
I think one major change here would be to avoid making `MLPMixer` as a struct with a custom forward pass. What I recommend instead is to define a new patching layer (in a separate file) like
```julia
struct Patching{T<:Integer}
  patch::T
end

@functor Patching

function (p::Patching)(x)
  h, w, c, n = size(x)
  hp, wp = h ÷ p.patch, w ÷ p.patch
  xpatch = reshape(x, hp, p.patch, wp, p.patch, c, n)

  return reshape(permutedims(xpatch, (1, 3, 5, 2, 4, 6), p.patch^2 * c, hp * wp, n)
end
```
Then `MLPMixer` is nothing more than a `Chain`. Part of the goal with this package's design is to illustrate how Flux's feature make it possible to build advanced models with minimal boilerplate. Reimplementing `Chain`-like forward passes is something we want to avoid when possible.

Overall, the implementation looks correct; we just need some design iterations. Great job!

src/vit-like/mlpmixer.jl
```diff
@@ -0,0 +1,87 @@
+# Utility function for creating a residual block with LayerNorm before the residual connection
+residualprenorm(planes, fn) = SkipConnection(Chain(fn, LayerNorm(planes)), +)
+
+# Utility function for 1D convolution
+conv1d(inplanes, outplanes, activation) = Conv((1, ), inplanes => outplanes, activation)
```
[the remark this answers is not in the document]
theabhirath: They're slightly model specific, so I was a little concerned about that. There's a very similar utility function `prenorm` that's without the residual connection that's used in the ViT

src/Metalhead.jl
```diff
@@ -30,8 +33,18 @@ export  AlexNet,
...
+end
+
+# ViT-like models
+include("vit-like/mlpmixer.jl")
+
+export  MLPMixer
+
+# use Flux._big_show to pretty print large models
+for T in (:MLPMixer,)
+    @eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
```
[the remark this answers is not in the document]
theabhirath: Yeah the idea here was that then `show` function will most likely not be `_maybe_big_show` for the MLPMixer given that the API is different here, but with the patching layer that you've suggested I think that might not be needed anymore

src/vit-like/mlpmixer.jl
theabhirath: That's a notational gaffe 😅 Still getting accustomed to the Julia way of having the batch dimension last. I'll do it via the `mean(x; dims = ...)` function, I was trying the `reduce` out and must've forgotten to replace it

theabhirath:
> I think one major change here would be to avoid making `MLPMixer` as a struct with a custom forward pass. What I recommend instead is to define a new patching layer (in a separate file) like
> 
> ```julia
> struct Patching{T<:Integer}
>   patch::T
> end
> 
> @functor Patching
> 
> function (p::Patching)(x)
>   h, w, c, n = size(x)
>   hp, wp = h ÷ p.patch, w ÷ p.patch
>   xpatch = reshape(x, hp, p.patch, wp, p.patch, c, n)
> 
>   return reshape(permutedims(xpatch, (1, 3, 5, 2, 4, 6), p.patch^2 * c, hp * wp, n)
> end
> ```
> 
> Then `MLPMixer` is nothing more than a `Chain`. Part of the goal with this package's design is to illustrate how Flux's feature make it possible to build advanced models with minimal boilerplate. Reimplementing `Chain`-like forward passes is something we want to avoid when possible.
> 
Will do this, but if you don't mind can I keep the TensorCast dep? I will remove it if there's a clear reason, but `einops` notation has become so ubiquitous across model implementations in Python frameworks that I thought it would be a lot more intuitive for it to be the same way here

src/Metalhead.jl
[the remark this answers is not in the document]
darsnack: Yeah in this case we won't need to, but if we were to have a custom show for a specific model, then we'd probably put that code alongside the model.

src/vit-like/mlpmixer.jl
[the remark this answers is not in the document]
darsnack: In that case, prepend the names with `_` to indicate that they are helpers for this file.

darsnack:
Personally, I find the einsum notation more confusing than the plain Julia, but of course, that's just my subjective opinion.

Mainly, I'm hesitant to take on another dependency just to mimic PyTorch. Especially since it seems to only serve indexing operations that Julia does quite well with `reshape`. If it turns out that ViT can be made faster by using TensorCast, then that seems like a good reason to take on the dep. Maybe other contributors can weigh in (cc @ToucheSir).

ToucheSir:
There are a [few einsum-related packages](https://github.com/mcabbott/TensorCast.jl#elsewhere), so if we do adopt one it would ideally be relatively lightweight + performant (flexibility is not an issue since Python frameworks don't have much in their einsum implementations). Because that could take a minute, I vote to defer that discussion to the ViT PR since this one doesn't require any einsum ops. That will let us merge this one asap :)

darsnack:
Okay @theabhirath does that sound like a reasonable plan? Use the non-einsum implementation here for now. Since this model is mostly there, this will let us merge without much back and forth.

If we decide to take on an einsum dep in the ViT PR, we can update this model's code as well.

theabhirath:
That makes sense to me. I'll make the changes 👍🏽

src/vit-like/mlpmixer.jl
theabhirath: > The global average pooling can simply be done with `mean(x; dims = ...)`. Though I'm confused how the batch dimension ended up first in the current code.

How do I put this notation in a `Chain` though? Do I define a helper layer for this as well?

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,69 @@
...
+([reference](https://arxiv.org/pdf/2105.01601)).
+
+# Arguments
+  `planes`: Number of dimensions in the input and output.
+  `expansion_factor`: Determines the number of dimensions in the intermediate layer.
+  `activation`: Activation function to use.
+  `dropout`: Dropout rate.
+  `dense`: Type of dense layer to use in the feedforward block.
+"""
+function feedforward(planes, expansion_factor = 4, dropout = 0., dense = Dense)
```
darsnack, line 20: `mlpblock` instead? Is this different from ViT? If not, then it can go in `layers.jl` file too. Let's make the activation customizable then?

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,69 @@
...
+- `image_size`: Size of the input image.
+- `channels`: Number of channels in the input image.
+- `patch_size`: Size of each patch fed into the network.
+- `planes`: Number of dimensions in every layer after the patch expansion layer.
+- `depth`: Number of layers in the network.
+- `expansion_factor`: Determines the number of dimensions in the intermediate layers.
+- `dropout`: Dropout rate in the feedforward blocks.
+- `nclasses`: Number of classes in the output.
+"""
+function MLPMixer(; image_size = 256, channels = 3, patch_size = 16, planes = 512,
```
darsnack, line 46: Just to be consistent with VGG which has these input size switches too.

```suggestion
function MLPMixer(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = 16, planes = 512, 
```

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,69 @@
...
+  token_mix = _conv1d
+  channel_mix = Dense
+
+  layers = []
+  push!(layers, Patching(patch_size))
+  push!(layers, Dense((patch_size ^ 2) * channels, planes))
+  append!(layers, [Chain(_residualprenorm(planes, feedforward(num_patches, expansion_factor, 
+                                          dropout, token_mix)),
+                         _residualprenorm(planes, feedforward(planes, expansion_factor, dropout, 
+                                          channel_mix)),) for _ in 1:depth])
```
darsnack, line 60:
```suggestion
  append!(layers, [Chain(_residualprenorm(planes, feedforward(num_patches, expansion_factor, dropout, token_mix)),
                         _residualprenorm(planes, feedforward(planes, expansion_factor, dropout, channel_mix)),)
                   for _ in 1:depth])
```

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,69 @@
...
+- `dropout`: Dropout rate in the feedforward blocks.
+- `nclasses`: Number of classes in the output.
+"""
+function MLPMixer(; image_size = 256, channels = 3, patch_size = 16, planes = 512, 
+                    depth = 12, expansion_factor = 4, dropout = 0., nclasses = 1000)
+  @assert (image_size % patch_size) == 0 "image size must be divisible by patch size"
+  
+  num_patches = (image_size ÷ patch_size) ^ 2
+  token_mix = _conv1d
+  channel_mix = Dense
```
darsnack, line 52: Do you think these should be customizable? If not, why not pass them in directly to the functions below instead of assigning a variable?

darsnack (changes requested):
Almost done. A few minor changes and we can merge.

src/other/mlpmixer.jl
theabhirath: They're not meant to be explicitly customisable, in that it's not expected that a user will ever have to actually change them. However, I've written it in that fashion so that it's clear what the role of each of those functions is in case someone ever wants to experiment

src/other/mlpmixer.jl
darsnack: I think the more we can reduce code duplication, the better. I would adopt the ViT interface for this function, then just pass in `expansion_factor * planes` in the MLPMixer builder code.

theabhirath: That should be fine. I'll do that

src/other/mlpmixer.jl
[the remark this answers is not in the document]
theabhirath: Have just done this - the other changes are now incorporated too. And I've added the token/channel mixing option for the package devs

src/Metalhead.jl
```diff
@@ -5,33 +5,40 @@ using Flux: outputsize, Zygote
...
-        MobileNetv2, MobileNetv3
+        MobileNetv2, MobileNetv3,
+        MLPMixer
 
 # use Flux._big_show to pretty print large models
 for T in (:AlexNet, :VGG, :ResNet, :GoogLeNet, :Inception3, :SqueezeNet, :DenseNet, :ResNeXt, 
-          :MobileNetv2, :MobileNetv3)
-  @eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
+    :MobileNetv2, :MobileNetv3, :MLPMixer)
+@eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
```
darsnack, line 41:
```suggestion
          :MobileNetv2, :MobileNetv3, :MLPMixer)
  @eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
```

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,86 @@
...
+  append!(layers, [Chain(_residualprenorm(planes, mlpblock(num_patches, 
+                                          expansion_factor * num_patches, 
+                                          dropout, token_mix)),
+                         _residualprenorm(planes, mlpblock(planes, 
+                                          expansion_factor * planes, dropout, 
+                                          channel_mix)),) for _ in 1:depth])
+
+  classification_head = Chain(_seconddimmean, Dense(planes, nclasses))
+
+  return Chain(layers..., classification_head)
```
darsnack, line 50: To easily separate backbone and classifier. Needs `classifier` and `backbone` functions defined for `MLPMixer` too.

```suggestion
  return Chain(Chain(layers...), classification_head)
```

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,86 @@
...
+- nclasses: the number of classes in the output
+"""
+function MLPMixer(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = 16, planes = 512, 
+                  depth = 12, expansion_factor = 4, dropout = 0., pretrain = false, nclasses = 1000)
+                    
+  layers = mlpmixer(imsize; inchannels, patch_size, planes, depth, expansion_factor, dropout, 
+                    nclasses)
+  pretrain && loadpretrain!(layers, string("MLPMixer"))
+  MLPMixer(layers)
+end
```
darsnack, line 82: Shouldn't we have a separate method for `MLPMixer(; pretrain = false)` that doesn't have all the toggles available. I guess generally the whole repo needs some refactoring to check compatibility for pretrained models.

test/other.jl
```diff
@@ -0,0 +1,7 @@
+using Metalhead, Test
+using Flux
+
+@testset "MLPMixer" begin
+    @test size(MLPMixer()(rand(Float32, 256, 256, 3, 67))) == (1000, 67)
+    @test_skip gradtest(MLPMixer(), rand(Float32, 256, 256, 3, 67))
```
darsnack, line 6: The batch size needs to be small for CI to run.

```suggestion
    @test size(MLPMixer()(rand(Float32, 256, 256, 3, 2))) == (1000, 2)
    @test_skip gradtest(MLPMixer(), rand(Float32, 256, 256, 3, 2))
```

darsnack (changes requested):
Looks good! There were a few things I missed on the last pass, sorry.

darsnack:
Also, FYI for the future: use `git rebase` instead of `git merge` when pulling in upstream changes to your PR branch. `git merge` will show all the upstream changes in the diff, making it hard to parse what's actually changed vs upstream and what's just showing as changed even though it's the same.

src/other/mlpmixer.jl
theabhirath: Given that this is a pretty niche model anyways, I'm removing the `pretrain` option entirely for now - when there's a consensus on what to do for the other models we can always come back and fix this one

src/Metalhead.jl
```diff
@@ -5,33 +5,40 @@ using Flux: outputsize, Zygote
...
-        MobileNetv2, MobileNetv3
+        MobileNetv2, MobileNetv3,
+        MLPMixer
 
 # use Flux._big_show to pretty print large models
 for T in (:AlexNet, :VGG, :ResNet, :GoogLeNet, :Inception3, :SqueezeNet, :DenseNet, :ResNeXt, 
-          :MobileNetv2, :MobileNetv3)
-  @eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
+          :MobileNetv2, :MobileNetv3, :MLPMixer)
+@eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
```
darsnack, line 41:
```suggestion
  @eval Base.show(io::IO, ::MIME"text/plain", model::$T) = _maybe_big_show(io, model)
```

src/other/mlpmixer.jl
```diff
@@ -0,0 +1,87 @@
...
+                    nclasses)
+  MLPMixer(layers)
+end
+
+@functor MLPMixer
+
+(m::MLPMixer)(x) = m.layers(x)
+
+backbone(m::MLPMixer) = m.layers[1]
+classifier(m::MLPMixer) = m.layers[2:end]
```
darsnack, line 87:
```suggestion
classifier(m::MLPMixer) = m.layers[2]
```

darsnack (approved):

theabhirath:
Thank you so much! The formatting is still a pain 😅 Some sort of auto-formatter really needs to be around to ensure this kind of thing doesn't sneak under the radar

darsnack:
Yeah I agree. There's a PR for Flux on this. I'm just waiting for that to settle on a style choice which I can duplicate here.

PR #105, opened 2022-02-02
Implementation of base ViT model

theabhirath:
This is an implementation of the base ViT model as detailed in the paper [An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale](https://arxiv.org/abs/2010.11929). This is quite the finicky model to implement and while I'm fairly sure I've done it correctly I would appreciate someone proofread the code just in case 😅 There's additional deps in the form of 
1. TensorCast (once #103 gets merged then it should not be a problem) and 
2. LinearAlgebra. This is because of a utility function I had to write for batched matrix multiplication (I poached it off a StackOverflow answer, I hope that's not plagiarism). The function might not be the fastest tool for the job, though, so there's that additional thing to discuss too.

cc: @DhairyaLGandhi for encouraging me to start working on a ViT model 😄


src/vit-based/utilities.jl
```diff
@@ -0,0 +1,19 @@
+using LinearAlgebra: mul!
+
+chunk(A, k::Int; dim::Int = 1) = 
+    (selectdim(A, dim, i) for i in Iterators.partition(axes(A,dim), cld(size(A,dim), k)));
+
+pair(t) = (isa(t, Pair)) ? t : (t, t)
+
+function batchmul!(C, A, B)
+    for j in axes(A, 4), i in axes(A, 3)
+        @views mul!(C[:, :, i, j], A[:, :, i, j], B[:, :, i, j])
```
DhairyaLGandhi, line 10: This does not seem AD friendly

DhairyaLGandhi:
Great go! Really excited to see this come through. I was thinking Transformers.jl might be a nice home for this to keep things organized.

src/vit-based/utilities.jl
DhairyaLGandhi: NNlib has some batched ops.

ToucheSir: Good call, I think `batched_mul` should be updated to support >3D arrays.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,174 @@
...
+        Dense(hidden_planes, inplanes),
+        Dropout(dropout))
+end
+
+struct attention
+  heads
+  scale
+  qkvlayer
+  outlayer
+end
```
darsnack, line 29: Struct names should be capitalized in Julia by convention. But I'd suggest a similar redesign here as I did with MLPMixer. A lot of the functions, like `chunk` or `batchedmul` are trying to replicate operations across various heads. This is more easily expressed using `Parallel` without needing a custom forward pass.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,174 @@
...
+  Transformer(layers)
+end
+
+function (m::Transformer)(x)
+  for (attn, fn) in m.layers
+    x = attn(x)
+    x = fn(x)
+  end
+  return x
+end
```
darsnack, line 98: Similar comment here. This is just a `Chain`, so no need to make a new type.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,174 @@
...
+  drop = Dropout(emb_dropout)
+  transformer = Transformer(planes, depth, heads, headplanes, mlppanes, dropout)
+  mlp_head = Chain(LayerNorm(planes), Dense(planes, nclasses))
+
+  ViT(patch_height, patch_width, planes, patchplanes, pos_embedding, cls_token, drop, transformer,
+      pool, mlp_head)
+end
+
+function (m::ViT)(x)
+  x = @cast x[(p1, p2, c), (h, w), b] := x[(h, p1), (w, p2), c, b] p1 in 1:m.ph, p2 in 1:m.pw
```
darsnack, line 163: Can reuse the patching layer here

darsnack (changes requested):
This is a great start. Why don't we tackle iterating the transformer layers first before moving onto ViT itself?

src/vit-based/utilities.jl
ToucheSir: As a stop-gap, perhaps tensorcast?

darsnack: Is `batched_mul` even needed if MHA is implemented using `Parallel`? I suppose there's a performance argument, but the current implementation isn't doing that.

darsnack: TensorCast appears to be doing a naive dot-product, I think:
```julia
julia> @pretty @reduce z[h, w, c, n] := sum(i) x[h, i, c, n] * y[i, w, c, n]
begin
    @boundscheck ndims(x) == 4 || throw(ArgumentError("expected a 4-tensor x[h, i, c, n]"))
    @boundscheck axes(x, 2) == axes(y, 1) || throw(DimensionMismatch("range of index i must agree"))
    @boundscheck axes(x, 3) == axes(y, 3) || throw(DimensionMismatch("range of index c must agree"))
    @boundscheck axes(x, 4) == axes(y, 4) || throw(DimensionMismatch("range of index n must agree"))
    @boundscheck ndims(y) == 4 || throw(ArgumentError("expected a 4-tensor y[i, w, c, n]"))
    local nightingale = transmute(x, Val((1, nothing, 3, 4, 2)))
    local waterbuffalo = transmute(y, Val((nothing, 2, 3, 4, 1)))
    z = dropdims(sum(@__dot__(nightingale * waterbuffalo), dims = 5), dims = 5)
end
```
@mcabbott can correct me if I wrote this operation improperly for TensorCast.

mcabbott: Yes that's correct. All `@reduce` knows how to do is to broadcast out & sum. Not always as terrible as it sounds but probably not the fastest.

The reason `batched_mul` doesn't yet support 4d arrays is being uncertain about how best to handle batch dims. At present it broadcasts e.g. `3,3,3 ⊠ 3,3,1 -> 3,3,3`, but if it wants to handle e.g. `3,3,3,3 ⊠ 3,3,1,3 -> 3,3,3,3` then it's not just a reshape to the 3d case.

CarloLucibello:
It would be good to have the attention layer and the whole transformer model in flux since they are of general use. Pytorch added them a while ago

src/vit-based/vit.jl
theabhirath: Yeah I was experimenting with a lot of things here before all the boilerplate was whittled away, so somehow that code just stayed 😂 I've trimmed it

src/vit-based/vit.jl
theabhirath: The problem is that `chunk` is not exactly across the heads so much as it is to split `q`, `k`, and `v` from each other. And while the subsequent reshaping operation is definitely something that can be done in parallel, the multiplication after that isn't symmetric - https://github.com/chengchingwen/Transformers.jl/blob/master/src/basic/mh_atten.jl, for example, is another place where the implementation is similar

src/vit-based/utilities.jl
theabhirath: Something like https://github.com/chengchingwen/Transformers.jl/blob/master/src/fix/batchedmul.jl might help, but it seems to be for 3D tensors still

src/vit-based/vit.jl
darsnack: True, and if this is part of Flux, then it is worth introducing a new type. I'll post a more complete suggestion later today.

theabhirath:
So this code has been cleaned up a _lot_ more - it looks much tighter IMO (I shifted a lot of the reusable layers to the `layers.jl` file as discussed). I think the main ViT model itself looks perfect - of course, if there's any design changes you think can be made, do let me know 😅. But the main issue now is that of the MH-Attention layer - it's still written in a fashion that is not AD-friendly and might definitely be slower than we want...

theabhirath:
> It would be good to have the attention layer and the whole transformer model in flux since they are of general use. PyTorch added them a while ago

I was trying to write this but then a look at a lot of PyTorch code, especially transformer-based ones made me realise that since vanilla attention is not really used as is in a lot of places, people end up having to write their own code anyways. There's a lot of fancy implementations of attention with changes in the MLP blocks, in the heads etc but writing a very general version and exposing a lot of options for the user will make it a little cluttered

src/vit-based/utilities.jl
ToucheSir: https://github.com/chengchingwen/Transformers.jl/blob/master/src/fix/batchedmul.jl#L13 appears to accept higher-dimensional inputs. I can't tell whether it's GPU friendly, however.

mcabbott: IIRC the code in Transformers might be an ancestor of that in NNlib? Both call CUDA routines, e.g. https://github.com/chengchingwen/Transformers.jl/blob/master/src/cuda/batch_gemm_gpu.jl, which I think only handle 3d arrays.

theabhirath: Exploring further led me to https://github.com/chengchingwen/NeuralAttentionlib.jl, which seems to have quite some stuff implemented - but it might mean taking on an additional dep

theabhirath: Unfortunately still seems to be 3D arrays🥲

theabhirath: Does https://github.com/chengchingwen/NeuralAttentionlib.jl/blob/master/src/matmul/matmul.jl fit the bill? I'm not familiar enough with the internals to comment 😅

src/layers.jl
```diff
@@ -108,19 +125,32 @@ end
 
 @functor Patching
 
-"""
-    mlpblock(planes, expansion_factor = 4, dropout = 0., dense = Dense)
+# Positional embedding layer used by many vision transformer-like models
+struct PosEmbedding
+  embedding_vector
+end
```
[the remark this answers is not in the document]
darsnack: A docstring here would be helpful too.

src/layers.jl
```diff
@@ -108,19 +125,32 @@ end
...
-"""
-function mlpblock(planes, hidden_planes, dropout = 0., dense = Dense; activation = gelu)
-  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
-        dense(hidden_planes, planes, activation), Dropout(dropout))
+@functor PosEmbedding
+
+# Class tokens used by many vision transformer-like models
+struct CLSTokens
+  cls_token
 end
```
darsnack, line 140:
```suggestion
struct CLSTokens{T}
  cls_token::T
end
```

+ docstring

src/layers.jl
```diff
@@ -108,19 +125,32 @@ end
...
+
+function(m::CLSTokens)(x)
+  cls_tokens = repeat(m.cls_token, 1, 1, size(x)[3])
+  x = cat(cls_tokens, x; dims = 2)
```
darsnack, line 144:
```suggestion
  return cat(cls_tokens, x; dims = 2)
```

src/layers.jl
```diff
@@ -108,19 +125,32 @@ end
...
+@functor CLSTokens
+
+# Utility function to decide if mean pooling happens inside the model
+struct CLSPooling
+  mode
+end
+
+(m::CLSPooling)(x) = (m.mode == "cls") ? x[:, 1, :] : _seconddimmean(x)
+
+@functor CLSPooling
```
darsnack, line 156: This doesn't need to happen every forward pass right? Instead of introducing a new type, we can put `x -> x[:, 1, :]` or `x -> _seconddimmean(x)` in the chain during the `vit` function.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,169 @@
...
+"""
+function Transformer(planes, depth, heads, headplanes, mlpplanes, dropout = 0.)
+  layers = [Chain(SkipConnection(prenorm(planes, MHAttention(planes; heads, headplanes, dropout)), +),
+                  SkipConnection(prenorm(planes, mlpblock(planes, mlpplanes, dropout)), +)) 
+            for _ in 1:depth]
+
+  Transformer(Chain(layers...))
+end
+
+(m::Transformer)(x) = m.layers(x)
```
darsnack, line 75: I would recommend getting rid of the `Transformer` type and turning it into a builder function. It's only a no-op wrapper around `Chain`.

darsnack (changes requested):
Awesome this is indeed looking quite a bit cleaner. I think the only major design consideration left is MHA as you mentioned. I haven't had the time to think about this, and I likely won't until this weekend has passed since I have a paper deadline on Sunday.

src/layers.jl
theabhirath: Yep, I've fixed the docstrings for the utility layers

src/layers.jl
theabhirath: I'm still relatively new to Julia so I'm just gonna express my amazement at how an anonymous function just integrates that neatly into the Chain API. That's just amazing🤯

theabhirath:
> Awesome this is indeed looking quite a bit cleaner. I think the only major design consideration left is MHA as you mentioned. I haven't had the time to think about this, and I likely won't until this weekend has passed since I have a paper deadline on Sunday.

No issues, I'll look up possible approaches that I can find in the meanwhile, good luck with your paper!

src/vit-based/utilities.jl
ToucheSir: @mcabbott based on the commit dates I think so. It wouldn't be too bad to just copy https://github.com/chengchingwen/Transformers.jl/blob/master/src/fix/batchedmul.jl#L13-L23 over for this PR and make it use `NNlib.batched_mul` instead. Keeping it as an internal function sidesteps the ambiguity question.

@theabhirath worth a try I'd say, though it's not clear to me whether those functions handle >3D inputs. If they do, we could consider adding it as a dep to get an MHA layer for free as well. JuliaHub tells me that only one additional transitive dep (PartialFunctions, which is relatively lightweight) would be added.

ToucheSir: I stand corrected, the MHA layer itself is relegated to [test code](https://github.com/chengchingwen/NeuralAttentionlib.jl/blob/f389adb71714f4245233e2e8e97292ebbac6148b/test/old_impl/mh_atten.jl#L35). However, the workhorse functions that power its implementation should be available as part of NeuralAttentionLib.

theabhirath: Yeah they're still for 3D inputs though, it's not going to help here 🥲 I'll try and see if I can find something else

darsnack:
I had a chance to implement a `Parallel`-based approach, which looks much cleaner in my opinion. It also appears to be faster than the PR `MHAttention` on my machine. We'd still want to test GPU stuff.

## `Parallel` implementation

```julia
using NNlib: batched_mul

struct Attention{T}
    qkv::T
end

Attention(in, out) = Attention(Dense(in, out * 3; bias = false))

@functor Attention

function (attn::Attention)(x::AbstractArray{T}) where T
    q, k, v = chunk(attn.qkv(x), 3; dim = 1)
    scale = convert(T, sqrt(size(q, 1)))
    score = softmax(batched_mul(batched_transpose(q), k) / scale)
    attention = batched_mul(v, score)

    return attention
end

struct MultiHead{T, S}
    heads::T
    projection::S
end

function MultiHead(in, out, nheads; dropout = 0.)
    inheads, outheads = chunk(1:in, nheads), chunk(1:out, nheads)
    heads = Parallel(vcat, [Attention(length(i), length(o)) for (i, o) in zip(inheads, outheads)]...)
    projection = Chain(Dense(out, out), Dropout(dropout))

    MultiHead(heads, projection)
end

@functor MultiHead

function (mha::MultiHead)(x)
    xhead = chunk(x, 3; dim = 1)

    return mha.projection(mha.heads(xhead...))
end
```

## Benchmarks

This is the code I used to setup the benchmark

```julia
dh = 64
nheads = 3
d, n, b = dh * nheads, 20, 32
x = rand(Float32, d, n, b)

mha = MultiHead(d, d, nheads)
mhapr = MHAttention(d; heads = nheads, headplanes = dh)
```

Here are the results

```julia
julia> @benchmark $(mha)($x)
BenchmarkTools.Trial: 303 samples with 1 evaluation.
 Range (min … max):  14.584 ms … 47.184 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     15.825 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.490 ms ±  2.918 ms  ┊ GC (mean ± σ):  0.43% ± 1.38%

   ▃▄▆█▄█▁▁   ▁                                                
  ▄████████▇▅██▄▄▄▅▅▅▃▃▃▁▂▃▂▁▃▁▁▂▁▁▁▁▁▁▁▂▂▂▂▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▂ ▃
  14.6 ms         Histogram: frequency by time        25.8 ms <

 Memory estimate: 5.26 MiB, allocs estimate: 438.

julia> @benchmark $(mhapr)($x)
BenchmarkTools.Trial: 85 samples with 1 evaluation.
 Range (min … max):  55.089 ms … 62.719 ms  ┊ GC (min … max): 0.00% … 3.66%
 Time  (median):     59.550 ms              ┊ GC (median):    3.79%
 Time  (mean ± σ):   59.256 ms ±  1.670 ms  ┊ GC (mean ± σ):  3.17% ± 1.71%

         ▁              ▁ ▁▃  ▁▁   ▃ ▁▆ ▃ ▃█▁                  
  ▄▁▁▁▄▄▄█▁▁▄▁▇▁▁▄▁▁▇▁▄▁█▄██▇▁██▄▇▇█▇██▇█▇███▁▇▁▁▇▄▇▁▁▁▄▄▁▁▄▄ ▁
  55.1 ms         Histogram: frequency by time        62.7 ms <

 Memory estimate: 50.50 MiB, allocs estimate: 5307.
```

## Some other notes

I think the implementation in the PR is wrong. It seems to be returning all zeros which I think is due to the last `@cast` operation.

The `chunk` utility here should be a PR to Flux or maybe renamed.

Perhaps we can look at speeding up `Parallel` separately from this PR.

darsnack:
A few more numbers with `(dh, h, n, b) == (64, 8, 100, 32)`.

```julia
julia> 
       @benchmark $(mha)($x)
BenchmarkTools.Trial: 24 samples with 1 evaluation.
 Range (min … max):  210.888 ms … 217.349 ms  ┊ GC (min … max): 0.35% … 0.42%
 Time  (median):     213.661 ms               ┊ GC (median):    0.35%
 Time  (mean ± σ):   213.556 ms ±   1.737 ms  ┊ GC (mean ± σ):  0.41% ± 0.13%

                █ █            ▃                                 
  ▇▁▁▇▇▇▁▁▁▁▇▁▁▁█▁█▁▁▁▁▁▁▁▁▇▇▁▇█▁▇▁▁▁▇▁▇▁▁▇▁▇▇▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▇ ▁
  211 ms           Histogram: frequency by time          217 ms <

 Memory estimate: 92.31 MiB, allocs estimate: 1164.

julia> 
       @benchmark $(mhapr)($x)
BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  1.881 s …  1.898 s  ┊ GC (min … max): 0.51% … 0.60%
 Time  (median):     1.886 s             ┊ GC (median):    0.51%
 Time  (mean ± σ):   1.888 s ± 8.711 ms  ┊ GC (mean ± σ):  0.54% ± 0.05%

  █               █                                      █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.88 s        Histogram: frequency by time         1.9 s <

 Memory estimate: 307.26 MiB, allocs estimate: 89820.
```

theabhirath:
> I think the implementation in the PR is wrong. It seems to be returning all zeros which I think is due to the last `@cast` operation.

Yeah there were quite some issues because I was trying to implement MHA in a rather more Pythonic way. `Parallel` seems to be a very nice Julian way to resolve the same problem - I made slight tweaks but overall your suggestion fits very well. It also solves the problem of taking on additional deps

> The chunk utility here should be a PR to Flux or maybe renamed.

https://github.com/FluxML/Flux.jl/issues/1841 😅

src/layers.jl
```diff
@@ -127,18 +189,32 @@ end
...
+Appends class tokens to the input that are used for classfication by many vision 
+transformer-like models. Instantiated with a class token vector which is a learnable parameter.
+"""
+struct CLSTokens{T}
+  cls_token::T
 end
+
+function(m::CLSTokens)(x)
+  cls_tokens = repeat(m.cls_token, 1, 1, size(x)[3])
+  return cat(cls_tokens, x; dims = 2)
```
mcabbott, line 217:
```suggestion
  cls_tokens = repeat(m.cls_token, 1, 1, size(x,3))
  return hcat(cls_tokens, x)
```

src/layers.jl
```diff
@@ -127,18 +189,32 @@ end
...
 
-Feedforward block used in many vision transformer-like models.
+Positional embedding layer used by many vision transformer-like models. Instantiated with an 
+embedding vector which is a learnable parameter.
+"""
+struct PosEmbedding{T}
+  embedding_vector::T
+end
+
+(p::PosEmbedding)(x) = x .+ p.embedding_vector[:, 1:size(x)[2], :]
```
mcabbott, line 201: Not in fact a vector. Can the field just be `array` or something? No need to repeat the struct's name.

Can the struct also be `PositionEmbedding` or something? "Pos" isn't saving much.

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+  q, k, v = chunk(attn.qkv(x), 3; dim = 1)
+  scale = convert(T, sqrt(size(q, 1)))
+  score = softmax(batched_mul(batched_transpose(q), k) / scale)
+  attention = batched_mul(v, score)
+
+  return attention
+end
+
+struct MHAttention{Q <: Integer, S, T}
+  nheads::Q
```
mcabbott, line 134: Is anything gained by allowing strange integers here? Otherwise:
```suggestion
struct MHAttention{S, T}
  nheads::Int
```

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+
+"""
+    Patching{T <: Integer}
+
+Patching layer used by many vision transformer-like models to split the input image into patches.
+Can be instantiated with a tuple `(patch_height, patch_width)` or a single value `patch_size`.
+"""
 struct Patching{T <: Integer}
   patch_height::T
   patch_width::T
```
mcabbott, line 175: Similarly: does this really need to be generic over strange integers? And I think the field names need not repeat the struct name:
```suggestion
struct Patching
  height::Int
  width::Int
```
And I wonder why "Patching" not just "Patch".

src/layers.jl
```diff
@@ -127,18 +189,32 @@ end
...
-function mlpblock(planes, hidden_planes, dropout = 0., dense = Dense; activation = gelu)
-  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
-        dense(hidden_planes, planes, activation), Dropout(dropout))
+    CLSTokens{T}
+
+Appends class tokens to the input that are used for classfication by many vision 
+transformer-like models. Instantiated with a class token vector which is a learnable parameter.
+"""
+struct CLSTokens{T}
+  cls_token::T
```
mcabbott, line 212: Again this isn't a vector. Although it could be, I think. 

Also, I think "CLS" is just "Class"? Not worth saving two letters:
```
struct ClassTokens{T}
    array::T
```

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+Feedforward block used in many vision transformer-like models.
+
+# Arguments
+- `planes`: Number of dimensions in the input and output.
+- `hidden_planes`: Number of dimensions in the intermediate layer.
+- `dropout`: Dropout rate.
+- `dense`: Type of dense layer to use in the feedforward block.
+- `activation`: Activation function to use.
+"""
+function mlpblock(planes, hidden_planes, dropout = 0., dense = Dense; activation = gelu)
```
mcabbott, line 105: What would be used in place of Dense here? 

I wonder if it should be:
```
mlpblock(planes::Int, hidden_planes::Int, activation = gelu; dropout::Real = 0.0)
```

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
-See also [`reluadd`](#).
+Self attention layer used by transformer models. Can be instantiated with a layer that produces
+the key, value and query vectors from the input.
 """
-addrelu(x, y) = @. relu(x + y)
+struct Attention{T}
+  qkv::T
+end
+
+Attention(in, out) = Attention(Dense(in, out * 3; bias = false))
```
mcabbott, line 120:
```suggestion
Attention(in::Integer, out::Integer) = Attention(Dense(in, out * 3; bias = false))
```

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+- `headplanes`: number of hidden channels per head
+- `mlppanes`: number of hidden channels in the MLP block
+- `dropout`: dropout rate
+"""
+function Transformer(planes, depth, heads, headplanes, mlpplanes, dropout = 0.)
+  layers = [Chain(SkipConnection(prenorm(planes, MHAttention(planes, headplanes, heads, dropout)), +),
+                  SkipConnection(prenorm(planes, mlpblock(planes, mlpplanes, dropout)), +)) 
+            for _ in 1:depth]
+
+  Chain(layers...)
```
mcabbott, line 23: On Flux master, this might be a good candidate for `Chain(layers)` to reduce load time.

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
-Useful as the `connection` argument for [`resnet`](#).
-See also [`addrelu`](#).
+# Arguments
+- `in`: Number of dimensions in the input.
+- `hidden`: Number of dimensions in the intermediate layer.
+- `nheads`: Number of attention heads.
+- `dropout`: Dropout rate for the projection layer.
 """
-reluadd(x, y) = @. relu(x) + relu(y)
+function MHAttention(in, hidden, nheads, dropout = 0.)
```
mcabbott, line 150: Won't this be ambiguous with the default type constructor? Could be more specific:

```suggestion
function MHAttention(in::Integer, hidden::Integer, nheads::Integer; dropout = 0.0)
```
I also wonder if dropout should be a keyword everywhere.

src/utilities.jl
```diff
@@ -1,6 +1,28 @@
+# Utility function for getting chunks of an ND-array along a particular dimension
+chunk(A, k::Int; dim::Int = 1) = 
+    (selectdim(A, dim, i) for i in Iterators.partition(axes(A,dim), cld(size(A,dim), k)));
```
mcabbott, line 3: It might be worth providing a gradient for this, to make one array like A, not 3.

src/layers.jl
darsnack: Or `mlpblock(in, hidden, activation = gelu; dense = Dense, dropout = 0.)` if there is a reason for customizing it.

src/layers.jl
darsnack: I would not store this field and use `length(mha.heads.layer)` instead. Just seems like a larger surface for error if someone could manually construct `MHAttention` with a mismatch here.

src/layers.jl
darsnack: That name is what I originally suggested. I thought it matched `Embedding`.

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+  project_out = !(nheads == 1 && hidden == in)
```
darsnack, line 151: If this is the case, then we should just return a `Attention`?

src/layers.jl
darsnack: Or just `token` for the field. I guess it is repetitive, but `array` is pretty vague IMO.

src/layers.jl
darsnack: It's a vector from the perspective of the paper, but in the practical code. Just like the patches are "vectors."

Maybe the name should be more specific since this type of positional embedding applies strictly to vision transformers? Perhaps `ViPostionEmbeddding` or `VisualPositionEmbedding` (that might be too long). The field can be called `signal` since we traditionally say "positional embedding signal."

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+- `mlpplanes`: number of hidden channels in the MLP block in the transformer
+- `headplanes`: number of hidden channels per head in the transformer
+- `dropout`: dropout rate
+- `emb_dropout`: dropout rate for the positional embedding layer
+- `pool`: pooling type, either "cls" or "avg"
+- `nclasses`: number of classes in the output
+"""
+function vit(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = (16, 16), planes = 1024, 
+  depth = 6, heads = 16, mlppanes = 2048, headplanes = 64, dropout = 0.1, emb_dropout = 0.1, 
+  pool = "cls", nclasses = 1000)
```
darsnack, line 51:
```suggestion
            depth = 6, heads = 16, mlppanes = 2048, headplanes = 64, 
            dropout = 0.1, emb_dropout = 0.1, pool = "cls", nclasses = 1000)
```

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+  Chain(layers...)
+end
+
+"""
+    vit(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = (16, 16), planes = 1024, 
+        depth = 6, heads = 16, mlppanes = 2048, headplanes = 64, dropout = 0.1, emb_dropout = 0.1, 
+        pool = "cls", nclasses = 1000)
+
+Creates a Vision Transformer model as detailed in the paper An Image is Worth 16x16 
+Words: Transformers for Image Recognition at Scale .
```
darsnack, line 32:
```suggestion
Creates a Vision Transformer (ViT) model.
```

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+  depth = 6, heads = 16, mlppanes = 2048, headplanes = 64, dropout = 0.1, emb_dropout = 0.1, 
+  pool = "cls", nclasses = 1000)
+
+  im_height, im_width = imsize
+  patch_height, patch_width = patch_size
+
+  @assert (im_height % patch_height == 0) && (im_width % patch_width == 0)
+  "Image dimensions must be divisible by the patch size."
+  @assert pool in ["cls", "avg"]
+  "Pool type must be either cls (cls token) or avg (mean pooling)"
```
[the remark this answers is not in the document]
darsnack: Agree that these should be symbols. An `ArgumentError` would be more correct here, though this usage is probably inconsistent in the other models too.

src/layers.jl
```diff
@@ -127,18 +189,32 @@ end
...
-        dense(hidden_planes, planes, activation), Dropout(dropout))
+    CLSTokens{T}
+
+Appends class tokens to the input that are used for classfication by many vision 
+transformer-like models. Instantiated with a class token vector which is a learnable parameter.
+"""
+struct CLSTokens{T}
+  cls_token::T
 end
+
```
darsnack, line 214:
```suggestion

ClassTokens(dim::Integer; init = Flux.zeros) = ClassTokens(init(1, 1, dim))

```

src/layers.jl
```diff
@@ -127,18 +189,32 @@ end
...
+    PosEmbedding{T}
 
-Feedforward block used in many vision transformer-like models.
+Positional embedding layer used by many vision transformer-like models. Instantiated with an 
+embedding vector which is a learnable parameter.
+"""
+struct PosEmbedding{T}
+  embedding_vector::T
+end
+
```
darsnack, line 200: There's no need to pad a singleton dimension with broadcasting.

```suggestion

PosEmbedding(embedsize, npatches; init = (dims) -> rand(Float32, dims)) =
  PosEmbedding(init(embedsize, npatches))

```

src/layers.jl
darsnack, line 201: This should be the correct size using the constructor above.

```suggestion
(p::PosEmbedding)(x) = x .+ p.embedding_vector
```

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+  
+  num_patches = (im_height ÷ patch_height) * (im_width ÷ patch_width)
+  patchplanes = inchannels * patch_height * patch_width
+
+  return Chain(Patching(patch_height, patch_width),
+               Dense(patchplanes, planes),
+               CLSTokens(rand(Float32, (planes, 1, 1))),
+               PosEmbedding(rand(Float32, (planes, num_patches + 1, 1))),
```
darsnack, line 67:
```suggestion
               ClassTokens(planes),
               PosEmbedding(planes, num_patches + 1),
```

src/layers.jl
darsnack: Perhaps we can be even more explicit here with `ClassTokenEmbedding`.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+
+# Arguments
+- `planes`: number of input channels
+- `depth`: number of layers
+- `heads`: number of attention heads
+- `headplanes`: number of hidden channels per head
+- `mlppanes`: number of hidden channels in the MLP block
+- `dropout`: dropout rate
+"""
+function Transformer(planes, depth, heads, headplanes, mlpplanes, dropout = 0.)
```
darsnack, line 18: This should be `transformer` since it isn't a constructor.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,116 @@
...
+  num_patches = (im_height ÷ patch_height) * (im_width ÷ patch_width)
+  patchplanes = inchannels * patch_height * patch_width
+
+  return Chain(Patching(patch_height, patch_width),
+               Dense(patchplanes, planes),
+               CLSTokens(rand(Float32, (planes, 1, 1))),
+               PosEmbedding(rand(Float32, (planes, num_patches + 1, 1))),
+               Dropout(emb_dropout),
+               Transformer(planes, depth, heads, headplanes, mlppanes, dropout),
+               (pool == "cls") ? x -> x[:, 1, :] : x -> _seconddimmean(x),
```
darsnack, line 70: Wrap the back bone in a `Chain` so that `backbone(vit)` is callable.

darsnack (changes requested):
Okay I think we're at the point of cleaning up the final API. In addition to Michael's comments, I've left my own notes.

src/vit-based/vit.jl
mcabbott: Should these outermost chains have names, like `Chain(backbone = Chain(...), ...)`?

src/vit-based/vit.jl
mcabbott: And probably with some types & kw:
```suggestion
function transformer(planes::Integer, depth::Integer, heads::Integer, headplanes::Integer, mlpplanes::Integer; dropout::Real = 0.0)
```

src/layers.jl
mcabbott: True, `PositionEmbedding` might be  too general.

src/vit-based/vit.jl
darsnack: Yes they should. Our original implementations were merged prior to that PR to Flux, but I have been planning on adding names for all the models in a separate PR.

src/layers.jl
darsnack: That's originally why I named it `MultiHead` in my code above so it could read `MultiHead(Attention, in, hidden, nheads)` eventually. But there's not really any other type of `Attention` that's used like this. There are other non-dot-product based attention mechanisms, but I've never seen them in MHA context. It's also not clear that they would match the same `Attention(i, o)` constructor.

src/layers.jl
mcabbott: Could be like `.patchvectors` or something, if this helps match the paper. Even `p.vectors` seems fine. Or `signal` as you say.

ToucheSir: Transformers.jl just calls the field `embedding`: https://github.com/chengchingwen/Transformers.jl/blob/a646572c5e05ab0e752df678e101bc0fbb7d92b3/src/basic/embeds/position_embed.jl#L8-L11

src/layers.jl
ToucheSir: I've seen this referred to as a "patch embedding" before, perhaps that resolves the awkwardness of "patching".

src/layers.jl
mcabbott: Oh I didn't see the original. It might be too weird to be generic without at least two cases in mind.

src/vit-based/vit.jl
ToucheSir: Nice that `layers` has a concrete eltype here.

src/vit-based/vit.jl
ToucheSir: Is there a more specific name for this than transformer? e.g. the ViT paper calls this part a "transformer encoder".

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+Attention(in, out) = Attention(Dense(in, out * 3; bias = false))
+
+@functor Attention
+
+function (attn::Attention)(x::AbstractArray{T}) where T
+  q, k, v = chunk(attn.qkv(x), 3; dim = 1)
+  scale = convert(T, sqrt(size(q, 1)))
+  score = softmax(batched_mul(batched_transpose(q), k) / scale)
+  attention = batched_mul(v, score)
```
mcabbott, line 128: This only accepts `x::AbstractArray{T,3}`, so I would write that. 

Many layers also accept input without the batch dimension. If this one wants to, it would need a separate implementation, without `batched_transpose`.

Also, does someone have a great answer for why `/ scale`, why not absorb this into the learnable parameters? I know everyone does this but am unsure whether it's a cargo cult.

src/layers.jl
```diff
@@ -91,28 +91,90 @@ end
...
+Can be instantiated with a tuple `(patch_height, patch_width)` or a single value `patch_size`.
+"""
 struct Patching{T <: Integer}
   patch_height::T
   patch_width::T
 end
+
 Patching(patch_size) = Patching(patch_size, patch_size)
 
 function (p::Patching)(x)
```
mcabbott, line 180: Should this be:
```suggestion
function (p::Patching)(x::AbstractArray{<:Number, 4})
```
Not sure, in fact, since I guess it will run with ndims < 4 too.

src/vit-based/vit.jl
darsnack: I think `transformer_encoder` is good considering only the rare user will call this directly

src/layers.jl
mcabbott: Should this constrain the type of `x`, if only as documentation?

src/layers.jl
theabhirath: For large values of `scale`, without the division, the softmax will have outputs that are very large, leading to small gradients which will impact the learning capacity of the model

mcabbott: This is what I've heard people say, but why is this better than scaling the weights that lead to `q`? In many implementations, that's a separate weight matrix.

darsnack: Without a constraint, I'm not sure `Wq` will naturally learn to scale down. Manually scaling down `Wq` in the forward pass will be tantamount to what's here. Unless you are asking why `scale` itself is not learned? I think the original attention paper mentions that it could be but `sqrt(d)` works well enough.

mcabbott: I guess I'm wondering why the scale isn't applied once at initialisation, instead of every pass. In this implementation, maybe that's tricky, since there is only one big weight matrix. But why don't implementations which aren't like that do it at init?

ToucheSir: IIRC this is a pretty common feature in transformer models, so let's not try to optimize it for this PR.

src/layers.jl
theabhirath: Making it `PatchEmbedding` - IMO that leaves the least room for confusion

src/layers.jl
theabhirath: > It's a vector from the perspective of the paper, but not in the practical code. Just like the patches are "vectors."
> 
> Maybe the name should be more specific since this type of positional embedding applies strictly to vision transformers? Perhaps `ViPostionEmbeddding` or `VisualPositionEmbedding` (that might be too long). The field can be called `signal` since we traditionally say "positional embedding signal."

I've made this `ViPosEmbedding`...which while adding some confusion should be fine simply because it's highly unlikely that anyone but devs will ever need to use this layer. Plus docstring is pretty clear about its functionality

src/layers.jl
theabhirath: Yeah keeping it general for now

src/layers.jl
```diff
@@ -91,31 +91,90 @@ end
...
+function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
+  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
+        dense(hidden_planes, planes, activation), Dropout(dropout))
+end
 
 """
-    reluadd(x, y)
+Self attention layer used by transformer models. Can be instantiated with a layer that produces
+the key, value and query vectors from the input.
+"""
```
darsnack, line 113:
```suggestion
    Attention(in => out)
    Attention(qkvlayer)

Self attention layer used by transformer models. Specify the `in` and `out` dimensions,
or directly provide a `qkvlayer` that maps an input the queries, keys, and values.
"""
```

src/layers.jl
```diff
@@ -91,31 +91,90 @@ end
...
+
+function (mha::MHAttention)(x)
+  nheads = length(mha.heads.layers)
+  xhead = chunk(x, nheads; dim = 1)
+  return mha.projection(mha.heads(xhead...))
+end
+
+"""
+Patch embedding layer used by many vision transformer-like models to split the input image into patches.
+Can be instantiated with a tuple of integers `(patch_height, patch_width)` or a single value `patch_size`.
```
darsnack, line 168:
```suggestion
    PatchEmbedding(patch_size)
    PatchEmbedding(patch_height, patch_width)

Patch embedding layer used by many vision transformer-like models to split the input image into patches.
```

src/layers.jl
```diff
@@ -124,21 +183,36 @@ function (p::Patching)(x)
                  hp * wp, n)
 end
 
-@functor Patching
+@functor PatchEmbedding
 
 """
-    mlpblock(planes, expansion_factor = 4, dropout = 0., dense = Dense)
+Positional embedding layer used by many vision transformer-like models. Instantiated with an 
+embedding vector which is a learnable parameter.
```
darsnack, line 190:
```suggestion
    ViPosEmbedding(embedsize, npatches; init = (dims) -> rand(Float32, dims))

Positional embedding layer used by many vision transformer-like models.
```

src/layers.jl
```diff
@@ -124,21 +183,36 @@ function (p::Patching)(x)
...
-  `hidden_planes`: Number of dimensions in the intermediate layer.
-  `dropout`: Dropout rate.
-  `dense`: Type of dense layer to use in the feedforward block.
-  `activation`: Activation function to use.
 """
-function mlpblock(planes, hidden_planes, dropout = 0., dense = Dense; activation = gelu)
-  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
-        dense(hidden_planes, planes, activation), Dropout(dropout))
+Appends class tokens to the input that are used for classfication by many vision 
+transformer-like models. Instantiated with a class token vector which is a learnable parameter.
```
darsnack, line 205:
```suggestion
    ClassTokens(dim; init = Flux.zeros)

Appends class tokens to an input with embedding dimension `dim` for use in many vision transformer models.
```

src/utilities.jl
darsnack: Also, if we are keeping this in Metalhead.jl, then it should be renamed so it doesn't conflict with `Flux.chunk`.

src/vit-based/vit.jl
darsnack: No need to spell out the title of the paper here.

src/vit-based/vit.jl
```diff
@@ -0,0 +1,115 @@
...
+  npatches = (im_height ÷ patch_height) * (im_width ÷ patch_width)
+  patchplanes = inchannels * patch_height * patch_width
+
+  return Chain(Chain(PatchEmbedding(patch_height, patch_width),
+                     Dense(patchplanes, planes),
+                     ClassTokens(planes),
+                     ViPosEmbedding(planes, npatches + 1),
+                     Dropout(emb_dropout),
+                     transformer_encoder(planes, depth, heads, headplanes, mlppanes, dropout),
+                     (pool == :class) ? x -> x[:, 1, :] : x -> _seconddimmean(x)),
```
darsnack, line 70:
```suggestion
                     (pool == :class) ? x -> x[:, 1, :] : _seconddimmean),
```

src/vit-based/vit.jl
```diff
@@ -0,0 +1,115 @@
...
+- `mlpplanes`: number of hidden channels in the MLP block in the transformer
+- `headplanes`: number of hidden channels per head in the transformer
+- `dropout`: dropout rate
+- `emb_dropout`: dropout rate for the positional embedding layer
+- `pool`: pooling type, either :class or :mean
+- `nclasses`: number of classes in the output
+"""
+function ViT(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = (16, 16), planes = 1024, 
+  depth = 6, heads = 16, mlppanes = 2048, headplanes = 64, dropout = 0.1, emb_dropout = 0.1, 
+  pool = :class, nclasses = 1000)
```
darsnack, line 102:
```suggestion
            depth = 6, heads = 16, mlppanes = 2048, headplanes = 64,
            dropout = 0.1, emb_dropout = 0.1, pool = :class, nclasses = 1000)
```

darsnack (changes requested):
Looks great, mostly docstrings that need updating.

src/utilities.jl
theabhirath: Made it `chunks` for the time being, until https://github.com/JuliaML/MLUtils.jl/pull/47 goes through and we can change to using that instead

src/layers.jl
mcabbott: Yes, not suggesting to deviate from the wisdom of the ancestors in this PR...

But this question aside, the other point about the batch dimension. The present signature accepts any array but only works for N=3, it should be restricted. And perhaps there ought to be a method for N=2 since it's simple:
```
function (attn::Attention)(x::AbstractMatrix{T}) where T
    q, k, v = chunk(attn.qkv(x), 3; dim = 1)
    scale = convert(T, sqrt(size(q, 1)))
    v * softmax(q' * k / scale)
end
```

src/layers.jl
```diff
@@ -91,31 +91,95 @@ end
...
-  patch_width::T
+struct MHAttention{S, T}
+  heads::S
+  projection::T
 end
-Patching(patch_size) = Patching(patch_size, patch_size)
 
-function (p::Patching)(x)
+"""
+    MHAttention(in, hidden, nheads, dropout = 0.)
```
mcabbott, line 140: This is correctly a keyword in the code, but not in docstring:
```suggestion
    MHAttention(in, hidden, nheads; dropout = 0.0)
```

src/layers.jl
```diff
@@ -124,21 +188,38 @@ function (p::Patching)(x)
...
-  `planes`: Number of dimensions in the input and output.
-  `hidden_planes`: Number of dimensions in the intermediate layer.
-  `dropout`: Dropout rate.
-  `dense`: Type of dense layer to use in the feedforward block.
-  `activation`: Activation function to use.
 """
-function mlpblock(planes, hidden_planes, dropout = 0., dense = Dense; activation = gelu)
-  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
-        dense(hidden_planes, planes, activation), Dropout(dropout))
+    ClassTokens(dim; init = Flux.zeros)
```
mcabbott, line 210:
```suggestion
    ClassTokens(dim; init = Flux.zeros32)
```
I think?

src/layers.jl
```diff
@@ -124,21 +188,38 @@ function (p::Patching)(x)
...
+
+Appends class tokens to an input with embedding dimension `dim` for use in many vision transformer models.
+"""
+struct ClassTokens{T}
+  token::T
 end
+
+ClassTokens(dim::Integer; init = Flux.zeros) = ClassTokens(init(dim, 1, 1))
```
mcabbott, line 218:
```suggestion
ClassTokens(dim::Integer; init = Flux.zeros32) = ClassTokens(init(dim, 1, 1))
```

src/layers.jl
```diff
@@ -91,31 +91,95 @@ end
...
+
+# Arguments
+- `planes`: Number of dimensions in the input and output.
+- `hidden_planes`: Number of dimensions in the intermediate layer.
+- `dropout`: Dropout rate.
+- `dense`: Type of dense layer to use in the feedforward block.
+- `activation`: Activation function to use.
 """
-addrelu(x, y) = @. relu(x + y)
+function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```
mcabbott, line 105: This still seems backwards. Two integers and a real number is a bit confusing, but integers and a function is standard elsewhere. All other `dropout`s are keywords. 
```suggestion
function mlpblock(planes::Integer, hidden_planes::Integer, activation = gelu; dense = Dense, dropout::Real = 0.0)
```
```suggestion
function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```
```suggestion
function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```
```suggestion
function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```
```suggestion
function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```
```suggestion
function mlpblock(planes, hidden_planes, dropout = 0.; dense = Dense, activation = gelu)
```

theabhirath: I kept it that way because it is very unlikely that the activation function changes, while dropout is tweaked regularly in usage - but if you think it's still preferred the other way, I'll change it

theabhirath: That's true. Then it makes sense for it to be a kwarg.

src/layers.jl
theabhirath: > But this question aside, the other point about the batch dimension. The present signature accepts any array but only works for N=3, it should be restricted. 

The latest version of the code has that 👍🏽

src/utilities.jl
```diff
@@ -1,5 +1,27 @@
+# Utility function for getting chunks of an ND-array along a particular dimension
+chunks(A, k::Int; dim::Int = 1) = 
+    (selectdim(A, dim, i) for i in Iterators.partition(axes(A,dim), cld(size(A,dim), k)));
```
CarloLucibello, line 3: I'm surprised the `gradtest` is passing for `Vit` since Zygote has troubles with this:
```julia
julia> chunks(A, k::Int; dim::Int = 1) = 
           (selectdim(A, dim, i) for i in Iterators.partition(axes(A,dim), cld(size(A,dim), k)));

julia> gradient(x -> sum(chunks(x, 2)[1]), rand(5))
ERROR: MethodError: no method matching getindex(::Base.Generator{Base.Iterators.PartitionIterator{Base.OneTo{Int64}}, var"#13#14"{Int64, Vector{Float64}}}, ::Int64)
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/Zygote/FPUm3/src/compiler/interface2.jl:0 [inlined]
  [2] _pullback(::Zygote.Context, ::typeof(getindex), ::Base.Generator{Base.Iterators.PartitionIterator{Base.OneTo{Int64}}, var"#13#14"{Int64, Vector{Float64}}}, ::Int64)
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface2.jl:9
  [3] _pullback
    @ ~/.julia/packages/Zygote/FPUm3/src/tools/builtins.jl:15 [inlined]
  [4] _pullback(::Zygote.Context, ::typeof(Zygote.literal_getindex), ::Base.Generator{Base.Iterators.PartitionIterator{Base.OneTo{Int64}}, var"#13#14"{Int64, Vector{Float64}}}, ::Val{1})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface2.jl:0
  [5] _pullback
    @ ./REPL[8]:1 [inlined]
  [6] _pullback(ctx::Zygote.Context, f::var"#15#16", args::Vector{Float64})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface2.jl:0
  [7] _pullback(f::Function, args::Vector{Float64})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface.jl:34
  [8] pullback(f::Function, args::Vector{Float64})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface.jl:40
  [9] gradient(f::Function, args::Vector{Float64})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/compiler/interface.jl:75
 [10] top-level scope
    @ REPL[8]:1
 [11] top-level scope
    @ ~/.julia/packages/CUDA/bki2w/src/initialization.jl:52
```
What's the explanation?

theabhirath: The `gradtest` is skipped for all models in Metalhead😅

CarloLucibello: Ah! Ok, we can address that in another PR

theabhirath: If it's the time/resources factor for CI being a problem, maybe we could add a message in the Contributions document asking for people to run `gradtest` locally. I'm doing that now just to check if the previous contributions are solid too. And I think the above failure means that this PR will have to wait for https://github.com/JuliaML/MLUtils.jl/pull/47 to be merged and Metalhead will have to take on MLUtils as a dep, right?

CarloLucibello: I don't think CI time would be a problem given that we just check that a single gradient doesn't error out, so we don't do expensive finite-difference comparisons etc... 

> And I think the above failure means that this PR will have to wait for https://github.com/JuliaML/MLUtils.jl/pull/47 to be merged and Metalhead will have to take on MLUtils as a dep, right?

yes we can go that way (and in any case Flux will depend on MLUtils as well, soonish). I just merged the PR and tagged a new MLUtils release, so here we should have a compat MLUtils >= 0.1.1

theabhirath: Screenshot 2022-02-10 at 1 14 54 PM
This is on an M1 Mac with 16 GB RAM...and while the tests pass, I am very surprised at the times taken for execution....is this normal? MobileNetv3 and DenseNet are especially surprising

theabhirath: > yes we can go that way (and in any case Flux will depend on MLUtils as well, soonish). I just merged the PR and tagged a new MLUtils release, so here we should have a compat MLUtils >= 0.1.1

Will make the change 👍🏽

theabhirath: Getting this error when I try to add MLUtils:

```
ERROR: The following 1 direct dependency failed to precompile:

MLUtils [f1d291b0-491e-4a28-83b9-f70985020b54]

Failed to precompile MLUtils [f1d291b0-491e-4a28-83b9-f70985020b54] to /Users/theabhirath/.julia/compiled/v1.8/MLUtils/jl_PwZka8.
ERROR: LoadError: cannot replace module Datasets during compilation
Stacktrace:
 [1] include(mod::Module, _path::String)
   @ Base ./Base.jl:421
 [2] include(x::String)
   @ MLUtils.Datasets ~/.julia/packages/MLUtils/2lblx/src/Datasets/Datasets.jl:1
 [3] top-level scope
   @ ~/.julia/packages/MLUtils/2lblx/src/Datasets/Datasets.jl:5
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:421
 [5] include(x::String)
   @ MLUtils ~/.julia/packages/MLUtils/2lblx/src/MLUtils.jl:1
 [6] top-level scope
   @ ~/.julia/packages/MLUtils/2lblx/src/MLUtils.jl:68
 [7] include
   @ ./Base.jl:421 [inlined]
 [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
   @ Base ./loading.jl:1399
 [9] top-level scope
   @ stdin:1
in expression starting at /Users/theabhirath/.julia/packages/MLUtils/2lblx/src/Datasets/datasets.jl:1
in expression starting at /Users/theabhirath/.julia/packages/MLUtils/2lblx/src/Datasets/Datasets.jl:1
in expression starting at /Users/theabhirath/.julia/packages/MLUtils/2lblx/src/MLUtils.jl:1
in expression starting at stdin:1
Stacktrace:
 [1] pkgerror(msg::String)
   @ Pkg.Types ~/julia/usr/share/julia/stdlib/v1.8/Pkg/src/Types.jl:67
 [2] precompile(ctx::Pkg.Types.Context, pkgs::Vector{String}; internal_call::Bool, strict::Bool, warn_loaded::Bool, already_instantiated::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Pkg.API ~/julia/usr/share/julia/stdlib/v1.8/Pkg/src/API.jl:1405
 [3] precompile
   @ ~/julia/usr/share/julia/stdlib/v1.8/Pkg/src/API.jl:1040 [inlined]
 [4] #precompile#225
   @ ~/julia/usr/share/julia/stdlib/v1.8/Pkg/src/API.jl:1037 [inlined]
 [5] precompile (repeats 2 times)
   @ ~/julia/usr/share/julia/stdlib/v1.8/Pkg/src/API.jl:1037 [inlined]
 [6] top-level scope
   @ REPL[3]:1
```

theabhirath: > weird, cannot reproduce this error (which I also don't understand, what's the problem with MLUtisl.Datasets?). You just add the package to Metalhead environment?

Yeah exactly...ironically, outside of the environment adding seems to be working just fine. Is the datasets module conflicting with something else by any chance?

> those timings are crazy, is that some serious precompilation problem for Zygote? what happens if you include("test/runtests.jl") twice?

I have set it to run, will update once I get the result

theabhirath: Screenshot 2022-02-10 at 3 59 08 PM
First and second runs. There's definitely serious first-compile issues here...the ViT `gradtest` is skipped for now until I resolve the MLUtils issue

CarloLucibello: that's crazy, better open a tracking issue.

I'll check the MLUtils issue

mcabbott: There is https://github.com/FluxML/Zygote.jl/issues/1119 . Flux master also has https://github.com/FluxML/Flux.jl/pull/1809 which may help some cases, and allows `Chain([Layer(), ...])` which helps some others.

mcabbott: The Flux changes will be 0.13 I think.

For the -O1 story, https://github.com/FluxML/Zygote.jl/pull/1147 saw some huge slowdowns elsewhere.

theabhirath: https://github.com/JuliaML/MLUtils.jl/issues/48#issue-1130216588 should solve the MLUtils problem

CarloLucibello (approved):

darsnack (approved):

darsnack:
Thank you for all the hard work and patience @theabhirath!

PR #109, opened 2022-02-05
Update pretrained artifacts

DhairyaLGandhi:
Now that we have some models trained, I wanted to start updating the Artifacts as well. I currently have resnets and mobilenets (I willl start ResNeXts and VGGs) and I believe @jeremiedb is also training some densenets. I wanted to open this now to clear out the which models need updating. 

We will need to make a release on Metalhead to get the weights all in one place as well, so I used a dummy repo to avoid making a release till we have the process figured out.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -14,6 +14,14 @@ lazy = true
     sha256 = "8ab8d60cc26e81451473badc9dc749b5ffc170a11bc00fb4b203da34fbfdc996"
     url = "https://github.com/FluxML/MetalheadWeights/releases/download/v0.1.1/googlenet-0.1.1.tar.gz"
 
+[resnet18]
+git-tree-sha1 = "e686115fea748bb63e6055076bbdfd6845ad07bb"
+lazy = true
+
+    [[resnet18.download]]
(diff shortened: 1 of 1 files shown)


darsnack:
As you upload the weights to the dummy repo, you can adjust https://github.com/FluxML/Metalhead.jl/blob/3997d1c2ce02162dc1d1ab21c5c6ab36448ba31a/test/convnets.jl#L4 in this PR. That will run the loading checks.

Are you wanting to do all the models in one release?

jbrea:
I wanted to try this PR with `model = ResNet(18, pretrain = true)` but I got an error message saying `Cannot locate artifact 'ResNet18'`. Then I changed in the `ResNet` function the string in `loadpretrain!` to `"resnet"`, which seemed to successfully download the artifact, but then I got a `Hash Mismatch!` error message, manipulated the Manifest.toml to fix this, but gave up eventually, when I hit `KeyError: key :weights not found`. Does it make sense to try loading pretrained weights starting with this PR or should I rather look at something like ONNX.jl?

darsnack:
The models in this PR now have pre-trained weights, and future additions should come as separate PRs using HuggingFace as specified in the contributing docs.

PR #111, opened 2022-02-07
Update first conv layer and default connection for ResNet

### LINKED ISSUE #110: ResNet50: difference between PyTorch and default Metalhead implementation
I noticed some differences between the PyTorch and the default Metalhead implementation for ResNet50.

* The `connection`defaults to  `connection = (x, y) -> @. relu(x) + relu(y)` while in PyTorch we have essentially `connection = (x, y) -> @. relu(x + y)` Unfortunately, the `ResNet50` constructor does not give access to the `connection` parameter.

https://github.com/pytorch/vision/blob/6db1569c89094cf23f3bc41f79275c45e9fcb3f3/torchvision/models/resnet.py#L119

* If we follow Pytorch, then the first `Conv((7,7))` layer should have no bias I believe:
https://github.com/pytorch/vision/blob/6db1569c89094cf23f3bc41f79275c45e9fcb3f3/torchvision/models/resnet.py#L146

Are there any particular reasons for this difference?

If we would follow the same structure than PyTorch, then we could (more easily) use the same weight as the PyTorch pretrained model. I have made a test with the PyTorch's ResNet50 and a custom Flux' ResNet50 model by reading all the parameters using the PyTorch model `state_dict()` method and this approach works quite well and I get exactly the same results from PyTorch than from Flux.

darsnack:
Fixes #110 by removing the bias from the first convolution layer and updating the default `connection` to `relu(x + y)`. Also exposes `connection` as part of the `ResNet` generic API.

### DIFF
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -36,7 +36,7 @@ function bottleneck(inplanes, outplanes, downsample = false)
 end
 
 """
-    resnet(block, residuals::NTuple{2, Any}, connection = (x, y) -> @. relu(x) + relu(y);
+    resnet(block, residuals::NTuple{2, Any}, connection = (x, y) -> @. relu(x + y);
            channel_config, block_config, nclasses = 1000)
 
 Create a ResNet model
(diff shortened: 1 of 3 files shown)


src/convnets/resnet.jl
```diff
@@ -135,17 +136,20 @@ See also [`resnet`](#).
...
+- `connection`: the binary function applied to the output of residual and skip paths in a block
 - `nclasses`: the number of output classes
 """
 struct ResNet
   layers
 end
 
-function ResNet(channel_config, block_config, shortcut_config; block, nclasses = 1000)
+function ResNet(channel_config, block_config, shortcut_config;
+                block, connection = (x, y) -> @. relu(x + y), nclasses = 1000)
```
ToucheSir, line 147: Good opportunity to extract `connection` as an internal helper function so that it (de)serializes better?

darsnack: `addrelu` and `reluadd`?

darsnack:
Is this good?

ToucheSir (approved):

PR #113, opened 2022-02-07
Implement EfficientNet

pxl-th:
This is an implementation of EfficientNet model in a similar way to [PyTorch EfficientNet](https://github.com/lukemelas/EfficientNet-PyTorch). 

### TODO

- [x] Documentation


pxl-th:
As for the pretrained weights, we can use [utilities](https://github.com/pxl-th/EfficientNet.jl/blob/master/src/load_utils.jl) to load PyTorch's weights to the model, save as `.bson` files and ship them as artifacts.

pxl-th:
Wanted to add pretrained weights, but looks like Metalhead's utilities for loading weights only include trainable parameters:
https://github.com/FluxML/Metalhead.jl/blob/4955fde24638d20ffbde3fe715b7fb91f9c53b25/src/utilities.jl#L23

But it is improtant to load moving mean and variance for the BatchNorm.
Otherwise accurancy of the model decreases significantly and does no match the PyTorch's version.
Maybe we should instead save the whole model directly, like [Flux documentation](https://github.com/FluxML/Metalhead.jl/blob/4955fde24638d20ffbde3fe715b7fb91f9c53b25/src/utilities.jl#L23) suggests?

pxl-th:
Additionally, maybe we should start adding inferrability tests for the models? (in light of [Taking TTFX seriously](https://discourse.julialang.org/t/taking-ttfx-seriously-can-we-make-common-packages-faster-to-load-and-use/74949))
However to make models that use BatchNorm type-stable we'd need a new release of Flux.jl with https://github.com/FluxML/Flux.jl/pull/1856.

ToucheSir:
> Maybe we should instead save the whole model directly, like [Flux documentation](https://github.com/FluxML/Metalhead.jl/blob/4955fde24638d20ffbde3fe715b7fb91f9c53b25/src/utilities.jl#L23) suggests?

Based on https://github.com/FluxML/MetalheadWeights/pull/2, that seems to be the way to go.

pxl-th:
Sorry, I don't have a lot of free time currently. It is totally ok if you are interested in completing this PR :)

darsnack:
Superseded by #171.

PR #114, opened 2022-02-08
ResNet: apply stride 2 to conv_bn((3,3)) as in PyTorch

Alexander-Barth:
While comparing the PyTorch and the Metalhead.ResNet(50/101/152) model, it seems to me that the stride=2 should be applied to the convolutional layer with kernel size 3.

https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py#L133

### PyTorch:
 ```
(layer3): Sequential(
    (0): Bottleneck(
      (conv1): Conv2d(512, 256, kernel_size=(1, 1), stride=(1, 1), bias=False)
      (bn1): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      # the following line is different
      (conv2): Conv2d(256, 256, kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), bias=False)  
      (bn2): BatchNorm2d(256, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      (conv3): Conv2d(256, 1024, kernel_size=(1, 1), stride=(1, 1), bias=False)
      (bn3): BatchNorm2d(1024, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      (relu): ReLU(inplace=True)
      (downsample): Sequential(
        (0): Conv2d(512, 1024, kernel_size=(1, 1), stride=(2, 2), bias=False)
        (1): BatchNorm2d(1024, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      )
    )
```

### Metalhead.ResNet(50):
```
Parallel(
        Metalhead.addrelu,
        Chain(
          Conv((1, 1), 512 => 256, stride=2, bias=false),  # 131_072 parameters
          BatchNorm(256, relu),         # 512 parameters, plus 512
          Conv((3, 3), 256 => 256, pad=1, bias=false),  # 589_824 parameters
          BatchNorm(256, relu),         # 512 parameters, plus 512
          Conv((1, 1), 256 => 1024, bias=false),  # 262_144 parameters
          BatchNorm(1024),              # 2_048 parameters, plus 2_048
        ),
        Chain(
          Conv((1, 1), 512 => 1024, stride=2, bias=false),  # 524_288 parameters
          BatchNorm(1024),              # 2_048 parameters, plus 2_048
        ),
      ),
```
With this change, I get the same results for both models (using the weights from PyTorch):

Here is for example the results of ResNet152:

```
resnet152
Flux:
acoustic guitar: 0.9387497
banjo: 0.01910968
electric guitar: 0.017470127
stage: 0.011316595
pick: 0.008845404
PyTorch:
acoustic guitar: 0.9387497
banjo: 0.019109717
electric guitar: 0.017470127
stage: 0.011316595
pick: 0.008845378
```

### DIFF
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -47,7 +47,7 @@ Create a ResNet model
            a new residual block (see [`Metalhead.basicblock`](#) and [`Metalhead.bottleneck`](#))
 - `residuals`: a 2-tuple of functions with input `(inplanes, outplanes, downsample=false)`,
                each of which will return a function that will be used as a new "skip" path to match a residual block.
-              [`Metalhead.skip_identity`](#) and [`Metalhead.skip_projection`](#) can be used here. 
+              [`Metalhead.skip_identity`](#) and [`Metalhead.skip_projection`](#) can be used here.
 - `connection`: the binary function applied to the output of residual and skip paths in a block
 - `channel_config`: the growth rate of the output feature maps within a residual block
 - `block_config`: a list of the number of residual blocks at each stage
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -30,8 +30,8 @@ Create a bottleneck residual block
 """
 function bottleneck(inplanes, outplanes, downsample = false)
   stride = downsample ? 2 : 1
-  Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = stride, bias = false)...,
-        conv_bn((3, 3), outplanes[1], outplanes[2]; stride = 1, pad = 1, bias = false)...,
+  Chain(conv_bn((1, 1), inplanes, outplanes[1]; stride = 1, bias = false)...,
+        conv_bn((3, 3), outplanes[1], outplanes[2]; stride = stride, pad = 1, bias = false)...,
         conv_bn((1, 1), outplanes[2], outplanes[3], identity; stride = 1, bias = false)...)
 end
 

### CODE REVIEW

**Review by darsnack — APPROVED** [review]
Thank you!

**Comment by Alexander-Barth** [comment]
I just found this note, in the PyTorch source code:

```
    # Bottleneck in torchvision places the stride for downsampling at 3x3 convolution(self.conv2)
    # while original implementation places the stride at the first 1x1 convolution(self.conv1)
    # according to "Deep residual learning for image recognition"https://arxiv.org/abs/1512.03385.
    # This variant is also known as ResNet V1.5 and improves accuracy according to
    # https://ngc.nvidia.com/catalog/model-scripts/nvidia:resnet_50_v1_5_for_pytorch.
```

https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py#L107

So this PR represents the ResNet v1.5 variant.

**Comment by darsnack** [comment]
Interesting, perhaps we can support both but use v1.5 by default? Though that gets into a territory where we make a judgement call about when to deviate from the literature. In this case, there is clearly a benefit, but what about #114 then?

**Comment by Alexander-Barth** [comment]
It is not so easy to strike a good balance between code readability and flexibility. For what it is worth I also checked the keras implementation and it is also using the v1.5 variant:
https://github.com/keras-team/keras/blob/d8fcb9d4d4dad45080ecfdd575483653028f8eda/keras/applications/resnet.py#L391

But I can make a PR to support both. Maybe with a version argument version = v"1.0" or v"1.5" (default)?

Additional discussion:
https://discuss.pytorch.org/t/resnet-50-v1-5-definition/34174/2

**Comment by darsnack** [comment]
Maybe the `bottleneck` block can accept a `strides` positional argument defaulting to `strides = [1, 2, 1]`? Then we can just link to the same Nvidia document that PyTorch does to indicate that we use the v1.5 version, and that to use the original, a user can specify `strides = [2, 1, 1]`?

It's not so pretty, but then someone can specify the `block` argument as a closure with the correct `strides`. I don't think most users will want to modify it, but if we get a lot of requests, then we can think about making it more convenient.

**Comment by darsnack** [comment]
Or perhaps even define
```julia
bottleneckv1(inplanes, outplanes, downsample = false) = bottleneck(inplanes, outplanes, downsample, [2, 1, 1])
```

PR #116, opened 2022-02-08
Update residual connection for ResNet18 and ResNet34

Alexander-Barth:
I noticed that for ResNet18 and ResNet34, PyTorch uses `skip_projection` (including a `Conv((1,))` and a `BatchNorm`) for the first parallel layer of all blocks except the first one (previously all parallel layers use `skip_identity` for these two configurations).

Note the `downsample` part is different:
```
[...]
(layer2): Sequential(
   (0): BasicBlock(
      (conv1): Conv2d(64, 128, kernel_size=(3, 3), stride=(2, 2), padding=(1, 1), bias=False)
      (bn1): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      (relu): ReLU(inplace=True)
      (conv2): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
      (bn2): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      (downsample): Sequential(
        (0): Conv2d(64, 128, kernel_size=(1, 1), stride=(2, 2), bias=False)
        (1): BatchNorm2d(128, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      )
 )
[...]
```
The current Metalhead has only a `MaxPool((1, 1), stride=2)` for downsampling.

```
 Parallel(                                                                                                                                                       
        Metalhead.addrelu,                                                                                                                                            
        Chain(                                                                                                                                                        
          Conv((3, 3), 64 => 128, pad=1, stride=2, bias=false),  # 73_728 parameters                                                                                  
          BatchNorm(128, relu),         # 256 parameters, plus 256                                                                                                    
          Conv((3, 3), 128 => 128, pad=1, bias=false),  # 147_456 parameters                                                                                          
          BatchNorm(128),               # 256 parameters, plus 256                                                                                                    
        ),                                                                                                                                                            
        Chain(MaxPool((1, 1), stride=2), #2),                                                                                                                         
      ),      
```

With this change, Metalhead.ResNet18 and Metalhead.ResNet34 are same as PyTorch.
~~If preferred, I can also change the PR to use two `residual` parameters for the `resnet` function: one `residual` parameter for the 1st layer (`firstresidual`) and one for the other layer (`residual`)~~. In the revised PR, `residual`and `shortcut_config` are vectors to allow for different type of connections for different blocks.

This should be all for the ResNet and VGG models.

### DIFF
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -212,7 +212,7 @@ resnet34 = ResNet([1, 1], [3, 4, 6, 3], :A; block = Metalhead.basicblock)
 ```
 
 The bottleneck of the orginal ResNet model has a stride of 2 on the first
-convolutional layer when downsampling (instead of the 2nd convolutional layers
+convolutional layer when downsampling (instead of the second convolutional layers
 as in ResNet v1.5). The architecture of the orignal ResNet model can be obtained
 as shown below:
 
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -108,12 +108,17 @@ function resnet(block, shortcut_config::AbstractVector{<:Symbol}, args...; kwarg
     :B => (skip_projection, skip_identity),
     :C => (skip_projection, skip_projection))
 
-  try
-    shortcut = [shortcut_dict[sc] for sc in shortcut_config]
-  catch
+  if any(sc -> !haskey(shortcut_dict,sc),shortcut_config)
     error("Unrecognized shortcut_config ($shortcut_config) passed to `resnet` (use a vector with :A, :B, or :C).")
   end
+
+  shortcut = [shortcut_dict[sc] for sc in shortcut_config]
   resnet(block, shortcut, args...; kwargs...)
+end
+
+function resnet(block, shortcut_config::Symbol, args...; block_config, kwargs...)
+    resnet(block, fill(shortcut_config, length(block_config)), args...;
+           block_config = block_config, kwargs...)
 end
 
 const resnet_config =
--- a/src/convnets/resnet.jl
+++ b/src/convnets/resnet.jl
@@ -53,7 +53,7 @@ Create a ResNet model
 - `block_config`: a list of the number of residual blocks at each stage
 - `nclasses`: the number of output classes
 """
-function resnet(block, residuals::NTuple{2, Any}, connection = addrelu;
+function resnet(block, residuals::AbstractVector{<:NTuple{2, Any}}, connection = addrelu;
                 channel_config, block_config, nclasses = 1000)
   inplanes = 64
   baseplanes = 64
@@ -61,24 +61,17 @@ function resnet(block, residuals::NTuple{2, Any}, connection = addrelu;
   append!(layers, conv_bn((7, 7), 3, inplanes; stride = 2, pad = 3, bias = false))
   push!(layers, MaxPool((3, 3), stride = (2, 2), pad = (1, 1)))
   for (i, nrepeats) in enumerate(block_config)
-    residuals_ =
-        if i == 1
-            residuals
-        else
-            (skip_projection, skip_identity)
-        end
-
     # output planes within a block
     outplanes = baseplanes .* channel_config
     # push first skip connection on using first residual
     # downsample the residual path if this is the first repetition of a block
     push!(layers, Parallel(connection, block(inplanes, outplanes, i != 1),
-                                       residuals_[1](inplanes, outplanes[end], i != 1)))
+                                       residuals[i][1](inplanes, outplanes[end], i != 1)))
     # push remaining skip connections on using second residual
     inplanes = outplanes[end]
     for _ in 2:nrepeats
       push!(layers, Parallel(connection, block(inplanes, outplanes, false),
-                                         residuals_[2](inplanes, outplanes[end], false)))
+                                         residuals[i][2](inplanes, outplanes[end], false)))
       inplanes = outplanes[end]
     end
     # next set of output plane base is doubled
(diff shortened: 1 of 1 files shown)


darsnack:
This difference is intentional. The PyTorch implementation is inconsistent with the original paper (see below). I suspect this is because the PyTorch implementation wasn't written to allow all the variations in the paper, and it was easier to make their ResNet-18/34 misalign than adjust the code.

Screen Shot 2022-02-08 at 9 02 50 AM

darsnack:
Is it possible to special-case (i.e. skip) `downsample` in the state dict in your script? If it's not too much difficulty, I'd rather do that and be consistent with the paper.

Alexander-Barth:
It is rather the other way around.
In PyTorch, I see 3 `skip_projection` for the 2nd, 3rd and 4th block (for the 1st `Parallel` layer in each block). Only the 1st block has a `skip_identity`. The 2nd `Parallel` layer in each block is always `skip_identity`.
If I output `residual_` for every block, it get this:

```
residuals_ = (Metalhead.skip_identity, Metalhead.skip_identity)     # block 1                                                                            
residuals_ = (Metalhead.skip_projection, Metalhead.skip_identity)   # block 2                                                                                                  
residuals_ = (Metalhead.skip_projection, Metalhead.skip_identity)   # block 3                                                                                                 
residuals_ = (Metalhead.skip_projection, Metalhead.skip_identity)   # block 4
```

ToucheSir:
Is there some documentation on the design choices made for the PyTorch implementation?

Alexander-Barth:
The documentation is really brief:
https://pytorch.org/vision/stable/models.html#torchvision.models.resnet18

Interestingly, tensorflow/keras do not provide ResNet18 or ResNet34. 

I agree with @darsnack that being able to use PyTorch pretrained weight is a big advantage and adding the original configuration from the paper as an option would indeed be a good compromise.

Alexander-Barth:
If you agree with these changes, I will also update the documentation. But if you want to keep the original implementation of ResNet18 and ResNet34, this is fine too (in this case we can close the PR).

src/convnets/resnet.jl
```diff
@@ -102,25 +102,31 @@ Create a ResNet model
...
+
+  if any(sc -> !haskey(shortcut_dict,sc),shortcut_config)
+    error("Unrecognized shortcut_config ($shortcut_config) passed to `resnet` (use a vector with :A, :B, or :C).")
   end
+
+  shortcut = [shortcut_dict[sc] for sc in shortcut_config]
   resnet(block, shortcut, args...; kwargs...)
 end
 
+function resnet(block, shortcut_config::Symbol, args...; block_config, kwargs...)
```
darsnack, line 119: Can we add a similar function for a single `NTuple` as well?

darsnack (changes requested):
The implementation looks good to me. Just the docstrings need to be updated.

src/convnets/resnet.jl
Alexander-Barth: Yes, done in my new commit.

src/convnets/resnet.jl
```diff
@@ -17,24 +17,31 @@ function basicblock(inplanes, outplanes, downsample = false)
...
+    bottleneck(inplanes, outplanes, downsample = false; stride = [1, (downsample ? 2 : 1), 1])
 
 Create a bottleneck residual block
-([reference](https://arxiv.org/abs/1512.03385v1)).
+([reference](https://arxiv.org/abs/1512.03385v1)). The bottleneck is composed of
+3 convolutional layers with the given `stride`. While the original paper uses a
+stride of 2 for the first convolutional layer (if `downsample` is `true`). A
+[revised variant](https://catalog.ngc.nvidia.com/orgs/nvidia/resources/resnet_50_v1_5_for_pytorch)
+called version 1.5 uses a stride of 2 (if `downsample` is `true`) in the 2nd
+convolutional layer instead.
```
darsnack, line 28:
```suggestion
([reference](https://arxiv.org/abs/1512.03385v1)). The bottleneck is composed of
3 convolutional layers each with the given `stride`.
By default, `stride` implements ["ResNet v1.5"](https://catalog.ngc.nvidia.com/orgs/nvidia/resources/resnet_50_v1_5_for_pytorch)
which uses `stride == [1, 2, 1]` when `downsample == true`.
This version is standard across various ML frameworks.
The original paper uses `stride == [2, 1, 1]` when `downsample == true` instead.
```

src/convnets/resnet.jl
```diff
@@ -102,25 +109,35 @@ Create a ResNet model
...
 end
 
+function resnet(block, shortcut_config::Symbol, args...; block_config, kwargs...)
+    resnet(block, fill(shortcut_config, length(block_config)), args...;
+           block_config = block_config, kwargs...)
+end
+
+function resnet(block, shortcut_config::NTuple{N,Symbol}, args...; kwargs...) where N
+    resnet(block, collect(shortcut_config), args...; kwargs...)
+end
```
darsnack, line 133: Sorry, I meant for the main builder method, not the one that looks up the shortcut config.

```suggestion
resnet(block, residuals::NTuple{2}, args...; kwargs...) = resnet(block, [residuals], args...; kwargs...)
```

src/convnets/resnet.jl
```diff
@@ -133,7 +150,9 @@ See also [`resnet`](#).
 # Arguments
 - `channel_config`: the growth rate of the output feature maps within a residual block
 - `block_config`: a list of the number of residual blocks at each stage
-- `shortcut_config`: the type of shortcut style (either `:A`, `:B`, or `:C`)
+- `shortcut_config`: the type of shortcut style (either `:A`, `:B`, or `:C`).
+   `shortcut_config` can also be a vector or tuple of symbols if different shortcut styles are applied to
```
darsnack, line 154:
```suggestion
   `shortcut_config` can also be a vector of symbols if different shortcut styles are applied to
```

src/convnets/resnet.jl
```diff
@@ -175,6 +197,35 @@ See also [`Metalhead.resnet`](#).
...
+(following the implementation in PyTorch). The published version of
+`ResNet(18)` and `ResNet(34)` used type `:A` shortcuts for all four blocks. The
+example below shows how to create a 18 or 34-layer `ResNet` using only type `:A`
+shortcuts:
+
+```julia
+using Metalhead
+
+resnet18 = ResNet([1, 1], [2, 2, 2, 2], :A;
+    block = Metalhead.basicblock)
```
darsnack, line 212:
```suggestion
resnet18 = ResNet([1, 1], [2, 2, 2, 2], :A; block = Metalhead.basicblock)
```

src/convnets/resnet.jl
```diff
@@ -175,6 +197,35 @@ See also [`Metalhead.resnet`](#).
...
+
+resnet34 = ResNet([1, 1], [3, 4, 6, 3], :A;
+    block = Metalhead.basicblock)
```
darsnack, line 215:
```suggestion
resnet34 = ResNet([1, 1], [3, 4, 6, 3], :A; block = Metalhead.basicblock)
```

src/convnets/resnet.jl
```diff
@@ -175,6 +197,35 @@ See also [`Metalhead.resnet`](#).
...
+```
+
+The bottleneck of the orginal ResNet model has a stride of 2 on the first
+convolutional layer when downsampling (instead of the 2nd convolutional layers
```
darsnack, line 219:
```suggestion
convolutional layer when downsampling (instead of the second convolutional layers
```

src/convnets/resnet.jl
```diff
@@ -175,6 +197,35 @@ See also [`Metalhead.resnet`](#).
...
+as in ResNet v1.5). The architecture of the orignal ResNet model can be obtained
+as shown below:
+
+```julia
+bottleneck_v1(inplanes, outplanes, downsample = false) =
+    Metalhead.bottleneck(inplanes, outplanes, downsample; stride = [(downsample ? 2 : 1), 1, 1])
```
darsnack, line 225: Can we just add a function like this to the code base? Then here we just say "use `Metalhead.bottleneck_v1` for the original behavior.

darsnack (changes requested):
Looks good! Just some minor tweaking and we can merge.

src/convnets/resnet.jl
Alexander-Barth: ok, no problem.

darsnack (approved):
Great, thanks @Alexander-Barth!

PR #118, opened 2022-02-14
Implement ESRGAN

rutvij-25:


src/other/esrgan.jl
```diff
@@ -0,0 +1,128 @@
+using Flux
+using Flux:@functor
+
+function ConvBlock(inc,out,k,s,p,use_act)
+    if use_act 
+        return Chain(
+            Conv((k,k),inc => out,stride = s,pad = p,bias=true),
```
CarloLucibello, line 7:
```suggestion
            Conv((k, k), inc => out, stride = s, pad = p, bias = true),
```
please check the use of whitespace throughout your code

CarloLucibello: thanks!

src/other/esrgan.jl
```diff
@@ -0,0 +1,132 @@
...
+
+function (m::DenseResidualBlock)(x) 
+    new_inputs = x
+    local out, new_inputs
+    for block in m.blocks
+        out = block(new_inputs)
+        new_inputs = cat(new_inputs, out, dims = 3)
+    end
+    return m.residual_beta * out + x
+end
```
ToucheSir, line 53: This can be replaced by a combination of built-in layers as suggested earlier. @darsnack should we consider adding a `BinaryMulAdd(A)` layer that does `(x, y) -> A .* x .+ b`?

src/other/esrgan.jl
```diff
@@ -0,0 +1,132 @@
...
+    conv = Conv((3, 3), nc => nc, stride = 1, pad = 1)
+    upsamples = Chain(UpsampleBlock(nc), UpsampleBlock(nc))
+    final = Chain(
+        Conv((3, 3), nc => nc, stride = 1, pad = 1, bias = true),
+        x -> leakyrelu.(x, 0.2),
+        Conv((3, 3), nc => inc, stride = 1, pad = 1, bias = true)
+    )
+
+    Generator(initial, residuals, conv, upsamples, final)
+end
```
ToucheSir, line 92: This should likewise get the built-in layer treatment. I'm unclear on what the purpose of separating the various components is, but if the names are very important then using the NamedTuple/keyword arg constructors for `Chain` should suffice.

src/other/esrgan.jl
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+    Discriminator(blocks, classifier)
+end
+
+function (m::Discriminator)(x)
+    x = m.blocks(x)
+    x = m.classifier(x)
+    return x
+end
```
ToucheSir, line 132:
```suggestion
function esrgan_discriminator(; in_c = 3, features = [64, 64, 128, 128, 256, 256, 512, 512])
    blocks = []
    for (idx, feature) in enumerate(features)
        s = 1 + (idx - 1) % 2
        push!(blocks, Conv((3, 3), in_c => feature, Base.Fix2(leakyrelu, 0.2), stride = s, pad = 1))
        in_c = feature
    end
    blocks = Chain(blocks...)
    classifier = Chain(
        AdaptiveMeanPool((6, 6)),
        flatten,
        Dense(512 * 6 * 6, 1024, Base.Fix2(leakyrelu, 0.2)),
        Dense(1024, 1)
    )

    return Chain(; blocks, classifier)
end
```
Can the 6 be adjusted? If it can, it should be a parameter.

ToucheSir (changes requested):
Thanks @rutvij-25, my comments below.

A couple of other points:
1.  Can you add tests for this?
2. I don't see an overall model struct like many of our models. This is trickier for GANs because of the split structure, but I would say to create one or both of a `ESRGAN` struct or `ESRGAN(Discriminator|Generator)` (need to think of a less verbose name here).

src/other/esrgan.jl
```diff
@@ -0,0 +1,132 @@
...
+
+@functor RRDB
+
+function RRDB(inc;
+              residual_beta = 0.2)
+    rrdb = Chain([DenseResidualBlock(inc) for _ in 1:3]...)
+    RRDB(residual_beta, rrdb)
+end
+
+(m::RRDB)(x) = m.rrdb(x) * m.residual_beta + x
```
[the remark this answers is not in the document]
rutvij-25: This is giving me an error
`no method matching Diagonal(::Float32, ::Float32)`

src/other/esrgan.jl
```diff
@@ -0,0 +1,132 @@
...
+        x -> leakyrelu.(x, 0.2),
+        Conv((3, 3), nc => inc, stride = 1, pad = 1, bias = true)
+    )
+
+    Generator(initial, residuals, conv, upsamples, final)
+end
+
+function (m::Generator)(x)
+    initial = m.initial(x)
+    x = m.conv(m.residuals(initial)) + initial
```
rutvij-25, line 96: I want to add 'initial' with the output of the next layer here, and I don't know how to do it just using Chain

theabhirath: Using [SkipConnection](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.SkipConnection) makes it:

`Chain(m.initial, SkipConnection(Chain(m.conv, m.residuals), +))`. Is that what you're looking for?

rutvij-25: I will check, and why are my builds failing :(( can you check it once! Thanks a ton

theabhirath: Well, there's a couple of things. ESRGAN doesn't really fit into the `show` function the way others do (because they're one model and also a single chain) so that's causing a precompilation error. Apart from that, `flatten` is exported by both MLUtils and Flux - this will be fixed by https://github.com/FluxML/Flux.jl/pull/1874 but you'll have to make it work by using `MLUtils.flatten` or `Flux.flatten` for now

src/other/esrgan.jl
```diff
@@ -0,0 +1,107 @@
...
+
+@functor esrgan_generator
+
+function (m::esrgan_generator)(x)
+    initial = m.initial(x)
+    x = m.conv(m.residuals(initial)) + initial
+    x = m.upsamples(x)
+    x = m.final(x)
+    return x
+end
```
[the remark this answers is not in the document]
rutvij-25: Got it!

src/other/esrgan.jl
```diff
@@ -0,0 +1,107 @@
...
+        Dense(1024, 1)
+    )
+    return Chain(blocks, classifier)
+end
+
+function ESRGAN()
+    return esrgan_discriminator(), esrgan_generator()
+end
+
+@functor ESRGAN
```
ToucheSir, line 107:
```suggestion
function esrgan()
    return Chain(generator=esrgan_discriminator(), discriminator=esrgan_generator())
end
```
One possible (but not necessarily optimal) way to do this. `@functor` on a normal function does nothing, generally speaking it only works on `struct`s.

src/other/esrgan.jl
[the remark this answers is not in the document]
ToucheSir: Looks like it's not exported by default, `Flux.Diagonal` should work.

src/other/esrgan.jl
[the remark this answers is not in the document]
darsnack: `ersgan` should probably return a `NamedTuple` containing the discriminator and generator (which are both functions like @ToucheSir mentioned). The `ERSGAN` wrapper struct can contain two fields for the generator and discriminator returned by `ersgan`. The wrapper struct option is weird for GANs, but I suppose it is nice for dispatching on the pair.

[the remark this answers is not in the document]
rutvij-25: @darsnack I have returned `Chain(generator=esrgan_discriminator(), discriminator=esrgan_generator())` instead of named tuple

rutvij-25:
need some help here! @darsnack

darsnack:
Sorry, I'll review this later today!

src/other/esrgan.jl
```diff
@@ -0,0 +1,86 @@
...
+        MLUtils.flatten,
+        Dense(512 * 6 * 6, 1024, Base.Fix2(leakyrelu, 0.2)),
+        Dense(1024, 1)
+    )
+    return Chain(blocks, classifier)
+end
+
+function esrgan()
+    return Chain(discriminator = esrgan_discriminator(), generator = esrgan_generator())
+end
```
darsnack, line 85: This should return a `NamedTuple` not a `Chain`. There is a separate loss for each term, so evaluating each piece explicitly makes more sense.

```suggestion
esrgan() = (generator = esrgan_generator(), discriminator = esrgan_discriminator())
```

darsnack (changes requested):
I left some comment mostly on the core functions. Try refactoring the outermost functions after incorporating these suggestions, then we can tackle reviewing those pieces.

PR #119, opened 2022-02-16
Implementation for ConvNeXt

theabhirath:
This is the implementation for [ConvNeXt](https://arxiv.org/abs/2201.03545). I've taken the liberty of refactoring the layers because a. a single file for `layers.jl` was getting way too cluttered and b. creating a Layers module might make it easier for reuse outside of the Metalhead repo without having to load the entire works. `gradtest` passes 🎊 so hopefully everything is fine


src/layers/drop.jl
```diff
@@ -0,0 +1,41 @@
+using Flux:Zygote
```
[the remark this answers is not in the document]
ToucheSir: About hard-coding `dims=4`, see my comment about layernorm and figuring out a way to support relative dims from the end upstream in Flux. For now I think it's fine to only work with 2D images.

src/convnets/convnext.jl
```diff
@@ -0,0 +1,91 @@
...
+                               Conv((2, 2), planes[m] => planes[m + 1]; stride = 2))
+      push!(layers, downsample_layer)
+    else
+      append!(layers, [ConvNeXtBlock(planes[m], dp_rates[cur + j], λ) for j in 1:depths[m]])
+      cur += depths[m]
+    end
+  end
+
+  head = Chain(GlobalMeanPool(),
+               x -> dropdims(x; dims = (1, 2)),
```
ToucheSir, line 62:
```suggestion
               MLUtils.flatten,
```
These are equivalent, no?

src/convnets/convnext.jl
```diff
@@ -0,0 +1,91 @@
...
+function convnext(; inchannels = 3, depths = [3, 3, 9, 3], planes = [96, 192, 384, 768], 
+                    drop_path_rate = 0.0f32, λ = 1fe-6, nclasses = 1000)
+  
+  layers = []
+  stem = Chain(Conv((4, 4), inchannels => planes[1]; stride = 4),
+               ChannelLayerNorm(planes[1]; ϵ = 1f-6))
+  push!(layers, stem)
+  dp_rates = LinRange{Float32}(0.0f32, drop_path_rate, sum(depths))
+  cur = 0
+  for i in 2:8
```
darsnack, line 49: Shouldn't this adjust with the length of `depths` and `planes`? I would recommend iterating `zip(depths, planes)`. The 2x iteration speed seems to be alternating between a downsampling layer and a sequence of blocks. Couldn't we just iterate at 1x and `push!` the downsampling layer then `append!` the blocks each iteration?

src/convnets/convnext.jl
```diff
@@ -0,0 +1,91 @@
...
+
+# Arguments:
+- `inchannels`: number of input channels.
+- `depths`: list with configuration for depth of each block
+- `planes`: list with configuration for number of output channels in each block
+- `drop_path_rate`: Stochastic depth rate.
+- `λ`: Init value for [LayerScale](https://arxiv.org/pdf/2103.17239)
+- `nclasses`: number of output classes
+"""
+function convnext(; inchannels = 3, depths = [3, 3, 9, 3], planes = [96, 192, 384, 768],
```
darsnack, line 40: Can we get a config dictionary for the various switches based on the paper?

darsnack (changes requested):
Looks awesome. This code is nearly ready out the gates. Most of the significant changes will be updating Flux layers to reduce code duplication. Otherwise, the code here looks good to me.

src/convnets/convnext.jl
theabhirath: Yep, I spent an hour figuring out all the dimensions and when the final work got put together that just got left in😅

src/convnets/convnext.jl
theabhirath: That is what I was working with initially.....I don't remember why I changed it now 😂 I'll make it that way

src/layers/normalise.jl
```diff
@@ -0,0 +1,65 @@
...
+end
+
+function(m::ChannelLayerNorm)(x)
+  x = MLUtils.normalise(x, dims = ndims(x) - 1, ϵ = m.ϵ)
+  m.diag === nothing ? m.λ.(x) : m.λ.(m.diag(x))
+end
+
+function ChannelLayerNorm(sz, λ = identity; affine = true, ϵ = 1f-5)
+  sz = sz isa Integer ? (sz,) : sz
+  diag = ChannelDiag(sz)
```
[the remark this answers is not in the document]
theabhirath: Yeah I just realised I wrote a lot of boilerplate for no reason 🤦🏽‍♂️ It would matter if the unsqueezed dims were greater than 1 because the number of learnable params would've changed but here this works just fine

src/layers/drop.jl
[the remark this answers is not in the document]
theabhirath: dims=4 is a really nifty trick....I missed that when I was searching because I was looking specifically for stochastic depth layers but that is exactly what's happening here so I think it's perfect

src/layers/normalise.jl
```diff
@@ -0,0 +1,65 @@
...
+struct ChannelLayerNorm{F,D,T,N}
+  λ::F
+  diag::D
+  ϵ::T
+  size::NTuple{N,Int}
+  affine::Bool
+end
+
+function(m::ChannelLayerNorm)(x)
+  x = MLUtils.normalise(x, dims = ndims(x) - 1, ϵ = m.ϵ)
```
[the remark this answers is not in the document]
theabhirath: Will do 👍🏽

theabhirath:
Most of the work is done - a Flux PR for the norm layers is pending, which I'll finalise in some time, and there's also the matter of weight initialisation. The paper follows truncated normal, which isn't in Flux yet - I'll try and put together a PR for that too. Is there also a single recursive method to initialise all the layers in a `Chain` in a specific way?

ToucheSir:
> Is there also a single recursive method to initialise all the layers in a `Chain` in a specific way?

Not that I'm aware of, since we don't have anything resembling a consistent interface for initialization across layers (even built-in ones).

src/layers/normalise.jl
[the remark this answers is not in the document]
theabhirath: On some reflection and experimentation I've come to realise that this is harder than I first thought, mostly because of dimension generalisations. In this particular case it's easy enough to pass the right constructor to `Diagonal`, but in case of random dimension choices ensuring the broadcasts work out and that the dimensions are maintained as is might be a problem. There's a very user-specific option which involves exposing the entire API to the user and asking them to decide the dimension choices and ensure that the size inputs to the `Diagonal` match, but I'm not sure that's the way to go for Flux...

src/convnets/convnext.jl
```diff
@@ -0,0 +1,104 @@
...
+([reference](https://arxiv.org/abs/2201.03545))
+
+# Arguments:
+- `planes`: number of input channels.
+- `drop_path_rate`: Stochastic depth rate.
+- `λ`: Init value for [LayerScale](https://arxiv.org/abs/2103.17239)
+"""
+function convnextblock(planes, drop_path_rate = 0., λ = 1f-6)
+  γ = λ > 0 ? Flux.ones32(planes) * λ : nothing
+  LayerScale(x) = isnothing(γ) ? x : x .* γ
```
mcabbott, line 14: What is this doing? Is the `γ` array meant to be trainable? I don't think `params` (or `functor`) will see it. 

If not, can't this just be `x -> x .* λ`?

theabhirath: Yep, it's intended to be learnable 😬 How do I make it visible to the `params` and `functor`?

mcabbott: I think it needs to be a struct, which you mark with `@functor`.

What you want is almost Flux.Diagonal, except without the β. It would be very easy to make `Flux.Diagonal(fill!(ones32(planes), λ), false)` work, but it doesn't right now as the signature is `Flux.Diagonal(::T, ::T)`.

theabhirath: > What you want is almost Flux.Diagonal, except without the β.

That's exactly what I want 😅 LayerScale is defined as 

"(it)...adds a learnable diagonal matrix on output of each residual block, initialised close to (but not at) 0. Adding this simple layer after each residual block improves the training dynamic, allowing for the training of deeper high-capacity image transformers that benefit from depth."

> it doesn't right now as the signature is `Flux.Diagonal(::T, ::T).`

Maybe a modification and an addition constructor to make the bias optional? I can open a PR if that's okay - this seems like something that would probably be useful in a lot of places

mcabbott: Yes, maybe this is spelled `Diagonal(init=λ, bias=false)` with `init::Number` understood as a scale?

Some chance others may prefer this to be a separate `Scale` layer --- the bias here can be the full size, not just a vector, but whether that makes it more or less like Dense is perhaps a matter of interpretation.

ToucheSir: In a sense `Dense` already handles this, but is somewhat poorly named (c.f. the example `Affine` layer in the docs or `Linear`). It also misses out on the fused broadcast, but that may be a non-issue given Zygote unfuses it anyhow.

darsnack: Not sure we have to special case `Number`, but a PR to Flux to make `Diagonal` accept the same initialization/hardcoded parameters in the constructor as everything else sounds like the right step to take here.

mcabbott: > Dense already handles this

What do you mean? This wants `w .* x` with some trainable array `w = fill(λ, ...)`, while Dense does `w * x`. 

> the same initialization/hardcoded parameters

But Diagonal doesn't initialise the bias the same way: It's not just a vector across channels, it's an array the same size as the other.

darsnack: Tbh, "diagonal" makes most sense to me. `Dense` is an affine transformation mathematically speaking. `Diagonal` is an affine transformation where the matrix is constrained to be diagonal.

> But Diagonal doesn't initialise the bias the same way: It's not just a vector across channels, it's an array the same size as the other.

Couldn't this be similar to `Dense` where `create_bias` is called with the correct dims? I agree it's not the same shape as the others, so we should not initialize it that way. I meant `Diagonal` should allow bias to be:
- the array itself
- `false`/`true`
Similarly, `Diagonal` should accept an `init` for the weight vector.

mcabbott: Re Dense, sure you could allow just a number, but here if I understood right we only start with a constant scale like that; afterwards each may be learned independently.

Seems a poor choice that Flux.Diagonal clashes with LinearAlgebra. Maybe we should even change it when doing these upgrades.

ToucheSir: It's an interesting thought, maybe in the future `Diagonal` could literally be `Dense` with a `Diagonal` weight matrix. What I haven't wrapped my head around is how all of this would function with >2D inputs, but that's not important for this PR.

theabhirath: If I've understood this correctly, then, I should open a PR for a modified version of Diagonal which makes the bias optional? Maybe the name can be discussed there further too

ToucheSir: No rush I think, we can get this PR through first.

theabhirath:
> Not that I'm aware of, since we don't have anything resembling a consistent interface for initialisation across layers (even built-in ones).

Given that the `trunc_norm` PR is slightly closer to merging, I have renewed interest in making this work somehow. All the built-in layers (well, at least the Dense and Conv families, which are the ones that are mostly of interest anyways) have an `init` keyword - the problem is that now there has to be a way to pass a value to `init` _after_ the layer has been initialised. One idea I had was to modify the initialisation functions to be mutating - then we could have a wrapper function that mutated the weights of the already existing layer. Not sure if this conflicts with something though

src/convnets/convnext.jl
theabhirath: Cool, then I'm leaving this part untouched for now. Some feedback on `ChannelLayerNorm` would be appreciated too, that one's been eating my head up far more than this

darsnack: This part still needs to be updated so that `params` captures the trainable diagonal. It would need to be changed to something like https://github.com/FluxML/Metalhead.jl/pull/119#discussion_r809195465 except instead of `false` you can use `Zeros()`. This would be what th

darsnack: This part still needs to be updated so that `params` captures the trainable diagonal. It would need to be changed to something like https://github.com/FluxML/Metalhead.jl/pull/119#discussion_r809195465 except instead of `false` you can use `Zeros()`. Having a convenience constructor for this can be done later than this PR.

src/layers/normalise.jl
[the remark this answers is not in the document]
darsnack: Yes, you're right including the `(1, 1, ...)` style changes will get hairy. I guess if we want this to be generic, `LayerNorm` should operate on a contiguous set of dimensions with a leading offset. We normalize the `offset:offset+length(sz)` dimensions of `x`, and we pad `ntuple(() -> 1, offset)` to the `Diagonal`. I don't think it is possible to make the broadcasting work out right for non-contiguous dimensions.

This seems like a complex change that would require a nicer user-facing API design too. If you want to tackle it as well, that would be great @theabhirath. But for this PR, I am fine with the current level of code duplication.

darsnack:
That sounds great! For the wrapper portion, I have some ideas about using Functors.jl for this. I would open up an issue on Flux.jl with your initial ideas (or PR if that makes more sense). We can continue the discussion there.

src/convnets/convnext.jl
theabhirath: I checked it out and the `show` function seems to be capturing the diagonal and it's parameters, and the number of total params right now check out with the original model, so somehow it's already working the way it's intended to

src/layers/normalise.jl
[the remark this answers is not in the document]
theabhirath: Yeah this is something I would definitely want to take up but I don't want to block this PR for that...I'll work on it separately and open a PR in Flux when I think I have something that can work well for this purpose

src/convnets/convnext.jl
```diff
@@ -10,8 +10,9 @@ Creates a single block of ConvNeXt.
 - `λ`: Init value for [LayerScale](https://arxiv.org/abs/2103.17239)
 """
 function convnextblock(planes, drop_path_rate = 0., λ = 1f-6)
-  γ = λ > 0 ? Flux.ones32(planes) * λ : nothing
-  LayerScale(x) = isnothing(γ) ? x : x .* γ
+  γ = Flux.ones32(planes) * λ
+  scale = λ > 0 ? identity : Diagonal(γ, Flux.Zeros())
+  LayerScale(x) = x .* scale
```
darsnack, line 15: Now `scale` can be passed into the `Chain` below. The forward pass of `Diagonal(γ, Flux.Zeros())` will do `x .* γ`.

darsnack: You're right I forgot about the matched types issue. Sorry, @theabhirath you'll need to revert to the `LayerScale` variation (but still keep the changes to avoid the `isnothing` check).

src/convnets/convnext.jl
```diff
@@ -0,0 +1,105 @@
...
+               Dense(planes[end], nclasses))
+  return Chain(layers..., head)
+end
+
+# Configurations for ConvNeXt models
+convnext_configs = Dict(:tiny => Dict("depths" => [3, 3, 9, 3], "planes" => [96, 192, 384, 768]),
+                        :small => Dict("depths" => [3, 3, 27, 3], "planes" => [96, 192, 384, 768]),
+                        :base => Dict("depths" => [3, 3, 27, 3], "planes" => [128, 256, 512, 1024]),
+                        :large => Dict("depths" => [3, 3, 27, 3], "planes" => [192, 384, 768, 1536]),
+                        :xlarge => Dict("depths" => [3, 3, 27, 3], "planes" => [256, 512, 1024, 2048]))
```
darsnack, line 75:
```suggestion
convnext_configs = Dict(:tiny => Dict(:depths => [3, 3, 9, 3], :planes => [96, 192, 384, 768]),
                        :small => Dict(:depths => [3, 3, 27, 3], :planes => [96, 192, 384, 768]),
                        :base => Dict(:depths => [3, 3, 27, 3], :planes => [128, 256, 512, 1024]),
                        :large => Dict(:depths => [3, 3, 27, 3], :planes => [192, 384, 768, 1536]),
                        :xlarge => Dict(:depths => [3, 3, 27, 3], :planes => [256, 512, 1024, 2048]))
```

src/convnets/convnext.jl
```diff
@@ -0,0 +1,105 @@
...
+
+# Arguments:
+- `inchannels`: number of input channels.
+- `drop_path_rate`: Stochastic depth rate.
+- `λ`: Init value for [LayerScale](https://arxiv.org/abs/2103.17239)
+- `nclasses`: number of output classes
+"""
+function ConvNeXt(mode::Symbol = :base; inchannels = 3, drop_path_rate = 0., λ = 1f-6, nclasses = 1000)
+  depths = convnext_configs[mode]["depths"]
+  planes = convnext_configs[mode]["planes"]
```
darsnack, line 95:
```suggestion
  depths = convnext_configs[mode][:depths]
  planes = convnext_configs[mode][:planes]
```

darsnack (approved):

darsnack:
Another day, another model. @theabhirath you're on a roll!

theabhirath:
> Another day, another model. @theabhirath you're on a roll!

Loving every second of it 😄 Onto the next one it is

PR #122, opened 2022-02-21
Minor fixes for ConNeXt

theabhirath:
This modifies some of the ConvNeXt code  - it makes the `backbone` function work as expected (earlier it wasn't because it wasn't wrapped in a `Chain`) and more significantly, cleans up `ChannelLayerNorm` quite a bit. After a lot of poking around, I figured that it isn't really used in a lot of papers (yet, at least), which is why the original paper actually writes this function from scratch. So I figured keeping it in Metalhead would be better but I've modified the docstring and the function itself to reflect the paper's implementation a lot more closely (it's cleaner too, as a result).

### DIFF
--- a/src/convnets/convnext.jl
+++ b/src/convnets/convnext.jl
@@ -58,7 +58,7 @@ function convnext(depths, planes; inchannels = 3, drop_path_rate = 0., λ = 1f-6
     cur += depths[i]
   end
 
-  backbone = collect(Iterators.flatten(Iterators.flatten((zip(downsample_layers, stages)))))
+  backbone = collect(Iterators.flatten(Iterators.flatten(zip(downsample_layers, stages))))
   head = Chain(GlobalMeanPool(),
                MLUtils.flatten,
                LayerNorm(planes[end]),
--- a/src/convnets/convnext.jl
+++ b/src/convnets/convnext.jl
@@ -58,13 +58,13 @@ function convnext(depths, planes; inchannels = 3, drop_path_rate = 0., λ = 1f-6
     cur += depths[i]
   end
 
-  layers = collect(Iterators.flatten(Iterators.flatten((zip(downsample_layers, stages)))))
-
+  backbone = collect(Iterators.flatten(Iterators.flatten((zip(downsample_layers, stages)))))
   head = Chain(GlobalMeanPool(),
                MLUtils.flatten,
                LayerNorm(planes[end]),
                Dense(planes[end], nclasses))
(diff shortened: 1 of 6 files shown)


theabhirath:
Tests failing on nightly for Windows and macOS because of `ERROR: LoadError: UndefVarError: libLLVMExtra not defined` for CUDA.jl. Upstream issue with Julia 1.9?

src/convnets/convnext.jl
```diff
@@ -58,13 +58,13 @@ function convnext(depths, planes; inchannels = 3, drop_path_rate = 0., λ = 1f-6
     cur += depths[i]
   end
 
-  layers = collect(Iterators.flatten(Iterators.flatten((zip(downsample_layers, stages)))))
-
+  backbone = collect(Iterators.flatten(Iterators.flatten((zip(downsample_layers, stages)))))
```
CarloLucibello, line 61:
```suggestion
  backbone = collect(Iterators.flatten(Iterators.flatten(zip(downsample_layers, stages))))
```
If my eyes don't trick me

src/layers/normalise.jl
```diff
@@ -1,30 +1,28 @@
+# Utility function for applying LayerNorm before a block
+prenorm(planes, fn) = Chain(fn, LayerNorm(planes))
```
CarloLucibello, line 2: If the block is `fn` layer norm is being applied afyer

theabhirath: Whoops 😅

src/layers/mlp.jl
```diff
@@ -10,7 +10,7 @@ Feedforward block used in many vision transformer-like models.
 - `dense`: Type of dense layer to use in the feedforward block.
 - `activation`: Activation function to use.
 """
-function mlpblock(planes, hidden_planes; dropout = 0., dense = Dense, activation = gelu)
+function MLPBlock(planes, hidden_planes; dropout = 0., dense = Dense, activation = gelu)
```
darsnack, line 13: This probably shouldn't get capitalized if it isn't a constructor.

darsnack (changes requested):
Sorry for the delay. Looks good though there are a couple extraneous changes?

src/layers/mlp.jl
theabhirath: Whoops. Changed that to `mlp_block` - wanted to make it more readable

darsnack (approved):

PR #123, opened 2022-02-22
Implementation of Res2Net and Res2NeXt

theabhirath:
This is an implementation for [Res2Net and Res2NeXt](https://arxiv.org/abs/1904.01169). The model itself is relatively clean, the bottleneck block...less so. The forward pass for this one is weird (for me to write in a Julian way, at least). Shape tests pass for now, but gradtests don't because of some concatenation operation, I think? (thought that would be a supported op, so a little confused if the error message is accurate or not)

For context, the bottleneck block looks like this:
Screenshot 2022-02-22 at 9 54 25 AM

I'm going to try to rewrite this myself in a cleaner way but some help would also be appreciated😅


theabhirath:
...that is the most surreal CI failure ever. `Received message KILL`?😳

theabhirath:
> That's probably from Github's machine. Probably too much memory is being consumed.

That shouldn't be OS dependent though, right?

PR #125, opened 2022-02-25
Implementation of ResMLP and gMLP (with improvements to MLPMixer and `PatchEmbedding`)

theabhirath:
This is an implementation of [ResMLP](https://arxiv.org/abs/2105.03404). In the process, I ended up doing quite a lot of things, including an almost complete rewrite of the base MLPMixer model itself to make it cleaner and more understandable, as well as by fixing stuff like `PatchEmbedding`, `DropPath` and `LayerScale`. There is also some utility function cleanup (and some random formatting errors that I fixed as I found them) but mostly this PR deals with MLPMixer, ResMLP and `PatchEmbedding`. `gradtest` passes, so I think this should be fine on that front

Edit: also added [gMLP](https://arxiv.org/abs/2105.08050) to the implementations


theabhirath:
What is with the Ubuntu failures with the KILL signal 🥲

ToucheSir:
Per one of Kyle's earlier comments, this may be more OOMs. I saw a similar issue pop up in FluxBench, so perhaps something is leaking in the test suite, something is being tested with too large of an input or GC is not running aggressively enough?

theabhirath:
> Per one of Kyle's earlier comments, this may be more OOMs. I saw a similar issue pop up in FluxBench, so perhaps something is leaking in the test suite, something is being tested with too large of an input or GC is not running aggressively enough?

Inputs are similar for all models, so probably unlikely? Not sure about GC or leakages in the test suite, although it's interesting that it seems to be OS dependent somehow. macOS has been unaffected across PRs, while Windows is fine on this one but errors on the res2net one

ToucheSir:
Right, but some of the newer models might be larger. I wonder if the glibc memory bloat issue might be a factor on linux, but this is all speculation.

theabhirath:
Huh. That's unexpected - I did `GC.gc()` between the groups of testsets (not individual models just yet). Is that Julia version dependent somehow?

darsnack:
The GC can certainly be version dependent. It's also possible that we are right on the edge of maxing out and the results are non-deterministic. You could try going more fine-grained or even add `GC.gc()` directly in `gradtest`'s definition.

As for why macOS never has these issues, it looks like those machines get twice as much memory as Windows or Linux: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

darsnack:
Sorry, edited my comment above after you pushed

theabhirath:
> it seems like the MLP mixer variants are more memory intensive

This is exactly what's confusing me - they're not. The ViT variants are the heaviest by _far_, and they didn't trip the memory when they were merged, so this is surprising to me because the MLPMixer variants are less intensive on both memory and compute

theabhirath:
But yeah, we will probably need a different approach for tests anyways given that some of the ViT variants will be in the multi-100Ms range - this seems hacky and also doesn't seem to work all the time

theabhirath:
Is this GTG?

src/convnets/convnext.jl
```diff
@@ -37,7 +34,7 @@ Creates the layers for a ConvNeXt model.
 - `λ`: Init value for [LayerScale](https://arxiv.org/abs/2103.17239)
 - `nclasses`: number of output classes
 """
-function convnext(depths, planes; inchannels = 3, drop_path_rate = 0., λ = 1f-6, nclasses = 1000)
+function convnext(depths, planes; inchannels = 3, drop_path_rate = 0., λ = 1e-6, nclasses = 1000)
```
darsnack, line 37: Why make these `Float64`?

src/convnets/convnext.jl
```diff
@@ -89,9 +86,12 @@ Creates a ConvNeXt model.
 - `drop_path_rate`: Stochastic depth rate.
 - `λ`: Init value for [LayerScale](https://arxiv.org/abs/2103.17239)
 - `nclasses`: number of output classes
+
+See also [`Metalhead.convnext`](#).
 """
-function ConvNeXt(mode::Symbol = :base; inchannels = 3, drop_path_rate = 0., λ = 1f-6, 
+function ConvNeXt(mode::Symbol = :base; inchannels = 3, drop_path_rate = 0., λ = 1e-6, 
                   nclasses = 1000)
+  @assert mode in keys(convnext_configs) "`size` must be one of $(keys(convnext_configs))"
```
darsnack, line 94:
```suggestion
  @assert mode in keys(convnext_configs) "`size` must be one of $(collect(keys(convnext_configs)))"
```

src/layers/mlp.jl
```diff
@@ -1,16 +1,47 @@
...
 - `hidden_planes`: Number of dimensions in the intermediate layer.
+- `outplanes`: Number of dimensions in the output - by default it is the same as `inplanes`.
 - `dropout`: Dropout rate.
-- `dense`: Type of dense layer to use in the feedforward block.
 - `activation`: Activation function to use.
 """
-function mlp_block(planes, hidden_planes; dropout = 0., dense = Dense, activation = gelu)
-  Chain(dense(planes, hidden_planes, activation), Dropout(dropout),
-        dense(hidden_planes, planes), Dropout(dropout))
+function mlp_block(inplanes, hidden_planes; outplanes = inplanes, dropout = 0., activation = gelu)
```
darsnack, line 13: Should `outplanes` be positional?

src/layers/embeddings.jl
```diff
@@ -1,27 +1,35 @@
...
 
-PatchEmbedding(patch_size) = PatchEmbedding(patch_size, patch_size)
+  im_height, im_width = imsize
+  patch_height, patch_width = patch_size
 
-function (p::PatchEmbedding)(x)
-  h, w, c, n = size(x)
-  hp, wp = h ÷ p.patch_height, w ÷ p.patch_width
-  xpatch = reshape(x, hp, p.patch_height, wp, p.patch_width, c, n)
+  norm_layer = typeof(norm_layer) != typeof(identity) ? norm_layer(embedplanes) : identity
```
darsnack, line 22: I would avoid this and make the keyword argument default to `planes -> identity`. And specify in the docstring that `norm_layer` should be a single arg constructor.

src/layers/others.jl
```diff
@@ -0,0 +1,38 @@
...
+"""
+    DropPath(p)
+
+Implements Stochastic Depth.
+([reference](https://arxiv.org/abs/1603.09382))
+
+# Arguments
+- `p`: rate of Stochastic Depth.
+"""
+DropPath(p) = p ≥ 0 ? Dropout(p; dims = 4) : identity
```
darsnack, line 38: Not a constructor, so should be lowercase.

src/layers/others.jl
```diff
@@ -0,0 +1,38 @@
...
+
+Implements LayerScale.
+([reference](https://arxiv.org/abs/2103.17239))
+
+# Arguments
+- `scale`: Scaling factor, a learnable diagonal matrix which is multiplied to the input.
+"""
+struct LayerScale
+  scale
+end
```
darsnack, line 12:
```suggestion
struct LayerScale{T<:AbstractVector{<:Real}}
  scale::T
end
```

src/layers/others.jl
```diff
@@ -0,0 +1,38 @@
...
+([reference](https://arxiv.org/abs/2103.17239))
+
+# Arguments
+- `planes`: Size of channel dimension in the input.
+- `λ`: initialisation value for the learnable diagonal matrix.
+"""
+LayerScale(planes::Int, λ) = λ > 0 ? LayerScale(Flux.ones32(planes) * λ) : identity
+
+@functor LayerScale
+(m::LayerScale)(x::AbstractArray{T}) where T = convert.(T, m.scale) .* x
```
darsnack, line 27: You don't need to manually convert here. You could enforce `Float32` in the constructor. Julia's promotion rules should be used on the forward pass.

src/other/mlpmixer.jl
```diff
@@ -1,83 +1,106 @@
...
 See also [`Metalhead.mlpmixer`](#).
 """
-function MLPMixer(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = 16, planes = 512, 
-                  depth = 12, expansion_factor = 4, dropout = 0., nclasses = 1000)
-                    
-  layers = mlpmixer(imsize; inchannels, patch_size, planes, depth, expansion_factor, dropout, 
-                    nclasses)
+function MLPMixer(size::Symbol = :base; patch_size::Int = 16, imsize::NTuple{2} = (224, 224),
+                  dropout = 0., drop_path_rate = 0., nclasses = 1000)
+  @assert size in keys(mixer_configs) "`size` must be one of $(keys(mixer_configs))"
```
darsnack, line 98:
```suggestion
  @assert size in keys(mixer_configs) "`size` must be one of $(collect(keys(mixer_configs)))"
```

src/layers/mlp.jl
```diff
@@ -1,16 +1,47 @@
...
+function gated_mlp(inplanes, hidden_planes; outplanes = inplanes, dropout = 0., activation = gelu, 
+									 gate_layer = identity)
+	layers = []
+	push!(layers, Dense(inplanes, hidden_planes, activation))
+	push!(layers, Dropout(dropout))
+	if typeof(gate_layer) != typeof(identity)
+		@assert hidden_planes % 2 == 0 "`hidden_planes` must be even for gated MLP"
+		gate_layer = gate_layer(hidden_planes)
+		hidden_planes ÷= 2
+	end
```
darsnack, line 42: Same suggestion here. Make `gate_layer` a 1-arg constructor, then you can just write out the `Chain` instead of pushing and splatting.

src/layers/embeddings.jl
```diff
@@ -1,27 +1,35 @@
...
 
-  return reshape(permutedims(xpatch, (1, 3, 5, 2, 4, 6)), p.patch_height * p.patch_width * c, 
-                 hp * wp, n)
+  @assert (im_height % patch_height == 0) && (im_width % patch_width == 0)
+  "Image dimensions must be divisible by the patch size."
+  
+  return Chain(Conv(patch_size, inchannels => embedplanes; stride = patch_size),
+               flatten ? x -> permutedims(reshape(x, (:, size(x, 3), size(x, 4))), (2, 1, 3)) 
+                       : identity,
+               norm_layer)
```
darsnack, line 30: This is quite different from before. The patches now have a learnable weight associated with them? That seems limiting to me. I would separate the learnable portion + normalization from the simple breaking up of an image into patches.

src/layers/embeddings.jl
```diff
@@ -1,27 +1,35 @@
...
+- `embedplanes`: the number of channels in the embedding
+- `norm_layer`: the normalization layer
+- `flatten`: whether to flatten the input after the embedding
 """
-struct PatchEmbedding
-  patch_height::Int
-  patch_width::Int
-end
+function PatchEmbedding(imsize::NTuple{2} = (224, 224); inchannels = 3, patch_size = (16, 16), 
+                        embedplanes = 768, norm_layer = identity, flatten = true)
```
darsnack, line 17: Where does 768 come from?

darsnack (changes requested):
Sorry about that, I left some comments. I'll need a little more time to review the spatial gating unit properly.

src/layers/embeddings.jl
theabhirath: Um, so this is the part that changed - the initial implementation wasn't a `PatchEmbedding`, it was just a Patching layer with a dense layer afterwards. But when I was seeing some more implementations, I realised that this learnable projection goes along with the patches every single time (for context, the timm implementations - from which the official repos of most recent papers borrow layers, use `PatchEmbed` in this fashion) - so it made sense to combine them in a format that's much more recognisable and also cleaner

src/layers/embeddings.jl
theabhirath: This is just a default value based on the 224-sized input, that's used in pretty much every paper without much changes (or with programmatically inserted values in case there are any changes) so...I just went with that 😅

src/layers/others.jl
theabhirath: Ah okay I'll make that change 👍🏽

src/layers/others.jl
theabhirath: This was more keeping the semantics of layers in mind: I'll make the change if it's a strict rule but given the usage I thought it would be more appropriate this way

src/layers/mlp.jl
theabhirath: In that case, do I move the `@assert` and the `hidden_planes` check + mutation to the inside of the constructor as well?

darsnack: Could it just be at the start of this function? Or is there a case with `identity` where `hidden_planes` isn't even? Otherwise we can just enforce it for all cases.

theabhirath: Yeah `hidden_planes` will probably be odd in some cases - it's calculated programmatically so can't really be sure about that. I'll move the checks to the start, but what about the `hidden_planes` mutation?

src/layers/mlp.jl
darsnack: Sure we can keep it as is. I just thought that it should be for consistency since all the other dimension arguments are.

theabhirath:
CI failures upstream on nightly during precompilation because `libblas not found` 🥲

src/layers/others.jl
theabhirath: > doesn't really meet the threshold a layer by Flux standards

Yeah this is why I kept it in Metalhead itself (same as for `ChannelLayerNorm`)  - I don't mind making it the direct version either, I just thought this helps Julian first-timers understand a little better how this works

src/layers/embeddings.jl
darsnack: Okay that makes sense. I suppose this should be more performant then. What about the normalization? It doesn't seem to be always used, and I see it kind of similar to batch norm. It's common to have conv + bn but we still keep the layers separate and provide a helper to `Chain` them. Maybe something like having the `PatchEmbedding` and `PatchNormalization` (which has the flatten + norm)?

Another note is that now this layer doesn't actually need to be a separate type...so naming issue again.

src/layers/embeddings.jl
darsnack: So is it calculated from 224 somehow? If it's calculated, then we should use the calculation formula as the default, not the result.

src/layers/mlp.jl
darsnack: Do you need to mutate? You could just pass `hidden_planes ÷ 2` to the next layer. I think you are correct that having the check inside the gate layer constructor is better than the start of the function.

Also, what about combining this with `mlp_block`? The `identity` case is the same as `mlp_block` right?

src/layers/others.jl
darsnack: Yeah but `ChannelLayerNorm` has a forward pass that isn't just a chain of existing layers (in this case it isn't even a `Chain` but convenience for a single keyword).

src/layers/mlp.jl
theabhirath: > You could just pass `hidden_planes ÷ 2` to the next layer.

Yes but this is subject to `gate_layer` not being identity

> what about combining this with `mlp_block`

This could be done, yeah - I kept it initially separate because of the checks

src/layers/embeddings.jl
theabhirath: The naming issues probably need to be decided on in a common format (because this is quite the common layer and camel-casing was my first thought 😅) but I'll default to the overall consensus. For the norm - `LayerNorm` has been used quite often, so that's probably one of the common choices, but this choice was written less from a current standpoint and more from the POV of people writing future models using these building blocks I reckon

src/layers/embeddings.jl
theabhirath: On searching a little more, I found that all papers seem to be exposing this as one of the fundamental hparams to tweak - case in point, ViTs:

image

(`embedplanes` follows from the timm implementation because it makes sense as the output of the `PatchEmbedding` layer but could always change that name to something more appropriate in case it's not informative enough)

src/other/mlpmixer.jl
```diff
@@ -87,3 +110,185 @@ end
...
+    SpatialGatingUnit(norm, proj)
+
+Creates a spatial gating unit as described in the gMLP paper.
+([reference](https://arxiv.org/abs/2105.08050))
+    
+# Arguments
+- `norm`: the normalisation layer to use
+- `proj`: the projection layer to use
+"""
+struct SpatialGatingUnit{T, F}
```
darsnack, line 194: Should this go in Layers?

test/other.jl
```diff
@@ -2,6 +2,34 @@ using Metalhead, Test
...
+@testset "gMLP" begin
+	@testset for mode in [:small, :base, :large, :huge]
+		@testset for drop_path_rate in [0.0, 0.5, 0.99]
+			m = gMLP(mode; drop_path_rate)
+			@test size(m(rand(Float32, 224, 224, 3, 2))) == (1000, 2)
+			@test_skip gradtest(m, rand(Float32, 224, 224, 3, 2))
+		end
+		GC.gc()
+	end
 end
```
darsnack, line 35:
```suggestion
end

```

darsnack (changes requested):
Wow this is a very clean implementation now that I've had the chance to appreciate it!

src/layers/others.jl
```diff
@@ -0,0 +1,38 @@
...
+    LayerScale(λ, planes::Int)
+
+Implements LayerScale.
+([reference](https://arxiv.org/abs/2103.17239))
+
+# Arguments
+- `planes`: Size of channel dimension in the input.
+- `λ`: initialisation value for the learnable diagonal matrix.
+"""
+LayerScale(planes::Int, λ) = λ > 0 ? LayerScale(Flux.ones32(planes) * λ) : identity
```
ToucheSir, line 24:
```suggestion
LayerScale(planes::Int, λ) = λ > 0 ? LayerScale(fill(Float32(λ), planes)) : identity
```

src/layers/mlp.jl
theabhirath: Yep this is perfect - dispatch in Julia is really a beast 😁

src/other/mlpmixer.jl
```diff
@@ -87,3 +110,185 @@ end
...
+
+# Arguments
+- `planes`: the number of planes in the block
+- `npatches`: the number of patches of the input
+- `norm_layer`: the normalisation layer to use
+"""
+function SpatialGatingUnit(planes::Int, npatches::Int; norm_layer = LayerNorm)
+  gateplanes = planes ÷ 2
+  norm = norm_layer(gateplanes)
+  proj = Dense(npatches, npatches)
```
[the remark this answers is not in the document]
theabhirath: Yep. This is pending because none of the models have a uniform weight initialisation interface, which is something I really want because it's a lot cleaner than hand-coding it in, so for now it's just left as is

src/other/mlpmixer.jl
theabhirath: This is niche enough that it's used only for this model, which is why I left it here for now - if I discover any future models that reuse it, we could always shift it there

src/other/mlpmixer.jl
```diff
@@ -1,83 +1,106 @@
...
-
-  return Chain(Chain(layers...), classification_head)
+- `embedplanes`: the number of channels after the patch embedding (denotes the hidden dimension)
+- `depth`: the number of blocks in the model
+- `mlp_ratio`: number(s) that determine(s) the number of hidden channels in the token mixing MLP 
+               and/or the channel mixing MLP as a ratio to the number of planes in the block.
+- `mlp_layer`: the MLP block to use
+- `dropout`: the dropout rate to use in the MLP blocks
+- `drop_path_rate`: Stochastic depth rate
+- `activation`: the activation function to use in the MLP blocks
```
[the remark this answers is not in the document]
theabhirath: This makes sense, but we lose default value information, do we not?

src/layers/others.jl
theabhirath: Will do, but this is going away soon enough after Flux 0.13 lands because Diagonal has this inbuilt now 😅

src/layers/others.jl
theabhirath: Yep, done now

theabhirath:
Why is Windows CI OOMing on nightly 😑 it's the same code for the stable version

theabhirath:
Yeah the memory issues are still problematic on GitHub Actions...probably need a long term alternative as the number of models increase

src/other/mlpmixer.jl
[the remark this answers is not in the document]
darsnack: We can still document the default set for the default value of `block` in the docstring. But as it stands, `block` is not really customizable unless it meets a strict set of arguments and keywords.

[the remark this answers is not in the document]
theabhirath: No I mean if we group the keywords how do we specify the default values?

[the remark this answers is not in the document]
darsnack: Is that not already handled by `mixerblock`? If `block_kwargs...` is empty meaning the user specified nothing, then you pass nothing to `mixerblock` which already has defaults.

This is not the case right now, but if `block` had a keyword argument with no default, but we are able to specify a default 1-level up, then we would use multiple dispatch of the type of `block` to insert that default into the `block_kwargs...`. Something like:
```julia
# use a shadow method
function _mlpmixer(block, ...; block_kwargs...)
 ...
end
mlpmixer(block::typeof(mixerblock), ...; specific_kw = val, kwargs...) = _mlpmixer(block, ...; specific_kw = val, kwargs...)
mlpmixer(block, ...; kwargs...) = _mlpmixer(block, ...; kwargs...)
```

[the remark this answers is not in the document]
theabhirath: This seems more like it - I'll work out a version of this that works with maximum API exposure if the user needs but hiding it completely in case they just want the default options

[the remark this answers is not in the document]
theabhirath: A complete version of this seems near impossible because the values are being used everywhere and exposing the API makes more sense, but I've cleaned it up a bit for now

src/layers/embeddings.jl
```diff
@@ -1,27 +1,34 @@
...
 
-PatchEmbedding(patch_size) = PatchEmbedding(patch_size, patch_size)
+# Arguments:
+- `imsize`: the size of the input image
+- `inchannels`: the number of channels in the input image
+- `patch_size`: the size of the patches
+- `embedplanes`: the number of channels in the embedding
+- `norm_layer`: the normalization layer - by default the identity function but otherwise takes a 
+                single argument constructor for a normalization layer like LayerNorm or BatchNorm
+- `flatten`: whether to flatten the input after the embedding
```
darsnack, line 15:
```suggestion
- `flatten`: set true to flatten the input spatial dimensions after the embedding
```

src/other/mlpmixer.jl
```diff
@@ -1,83 +1,100 @@
...
+"""
+    mlpmixer(block, imsize::NTuple{2} = (224, 224); inchannels = 3, norm_layer = LayerNorm,
+             patch_size::NTuple{2} = (16, 16), embedplanes = 512, drop_path_rate = 0.,
+             depth = 12, nclasses = 1000, kwargs...)
 
 Creates a model with the MLPMixer architecture.
 ([reference](https://arxiv.org/pdf/2105.01601)).
 
 # Arguments
+- `block`: the type of mixer block to use in the model - architecture dependent
```
darsnack, line 40:
```suggestion
- `block`: the type of mixer block to use in the model - architecture dependent
           (a constructor of the form `block(embedplanes, npatches; drop_path_rate, kwargs...)`)
```

src/other/mlpmixer.jl
```diff
@@ -87,3 +104,182 @@ end
...
+
+# Arguments
+- `planes`: the number of planes in the block
+- `npatches`: the number of patches of the input
+- `norm_layer`: the normalisation layer to use
+"""
+function SpatialGatingUnit(planes::Int, npatches::Int; norm_layer = LayerNorm)
+  gateplanes = planes ÷ 2
+  norm = norm_layer(gateplanes)
+  proj = Dense(Array{Float32}(undef, npatches, npatches), ones(npatches))
```
darsnack, line 207: Not sure what's a great value to use here, but plain `undef` could be arbitrarily large (as opposed to "close to zero" which is what we want).
```suggestion
  proj = Dense(2 * eps(Float32) .* rand(Float32, npatches, npatches), ones(npatches))
```

darsnack (changes requested):
Some small changes, but looks ready otherwise.

darsnack (approved):

darsnack:
Thanks @theabhirath!

PR #128, opened 2022-03-08
Let `cat_channels` accept more args

### LINKED ISSUE #127: `cat_channels` should support > 2 args
Flux's `Parallel` no longer performs a binary reduction. It calls the function on all paths at once. We should update `cat_channels` to support this case.

mcabbott:
Closes #127

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -28,13 +28,13 @@ See also [`addrelu`](#).
 reluadd(x, y) = @. relu(x) + relu(y)
 
 """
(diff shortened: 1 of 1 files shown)


darsnack (approved):

mcabbott:
Worth tagging so that Flux downstream tests work, perhaps?

darsnack:
Yes, but there are breaking changes on master, so I'll need to do due diligence and make sure everything is okay since we've had some churn on the newer APIs. I'll open a PR.

PR #130, opened 2022-03-11
Implementation of ConvMixer

theabhirath:
This is an implementation of [ConvMixer](https://arxiv.org/abs/2201.09792). This is a _pretty_ small model (to code, params weigh in at a solid 52M), but has some interesting experiments done around it. Something I cooked up in between trying to make some larger models work out...

### DIFF
--- a/src/convnets/convmixer.jl
+++ b/src/convnets/convmixer.jl
@@ -1,5 +1,5 @@
 """
-    convmixer(planes, depth; inchannels = 3, kernel_size = (9, 9), patch_size = 7,
+    convmixer(planes, depth; inchannels = 3, kernel_size = (9, 9), patch_size::Dims{2} = 7,
               activation = gelu, nclasses = 1000)
 
 Creates a ConvMixer model.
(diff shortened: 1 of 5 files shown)


theabhirath:
Bump?

src/convnets/convmixer.jl
```diff
@@ -0,0 +1,68 @@
...
+([reference](https://arxiv.org/abs/2201.09792))
+
+# Arguments
+- `planes`: number of planes in the output of each block
+- `depth`: number of layers
+- `inchannels`: number of channels in the input
+- `kernel_size`: kernel size of the convolutional layers
+- `patch_size`: size of the patches
+- `activation`: activation function used after the convolutional layers
+- `nclasses`: number of classes in the output
```
CarloLucibello, line 48: these do not correspond to the signature

src/layers/conv.jl
```diff
@@ -34,6 +35,10 @@ function conv_bn(kernelsize, inplanes, outplanes, activation = relu;
     bnplanes = outplanes
   end
 
+  if preact
+    activations = (conv = activation, bn = identity)
+  end
```
CarloLucibello, line 40: are all 4 combinations of the `rev` and `preact` working as they are supposed to? just checking

CarloLucibello:
needs a mention of ConvMixer in the docs (which is just the README?)

src/convnets/convmixer.jl
theabhirath: Fixed now!

src/layers/conv.jl
theabhirath: There was one edge case not being handled, I think - should be working now

src/layers/conv.jl
```diff
@@ -12,6 +12,7 @@ Create a convolution + batch normalization pair with ReLU activation.
 - `outplanes`: number of output feature maps
 - `activation`: the activation function for the final layer
 - `rev`: set to `true` to place the batch norm before the convolution
+- `preact`: set to `true` to place the activation function before the batch norm
```
darsnack, line 15:
```suggestion
- `preact`: set to `true` to place the activation function before the batch norm
            (only compatible with `rev = false`)
```

darsnack (changes requested):
One small suggestion. Looks good otherwise!

darsnack (approved):

PR #134, opened 2022-03-15
Swin Transformer related utility function

XWHtorrentx:
add a function which returns a matrix of relative position index from Swin Transformer

### DIFF
--- a/src/layers/relative_index.jl
+++ b/src/layers/relative_index.jl
@@ -1,0 +1,32 @@
+#Utility function for obtaining relative index in shifted window transformer
+#Created by Weihang Xia
+using Flux
+using Base.Threads
+#This function returns a grid of tensor x and y
+function meshgrid(x,y)
+    xv=first.(Iterators.product(x,y))
+    yv=last.(Iterators.product(x,y))
+    return [xv,yv]
+end
+
+#Input is a 2-Tuple of window_size, the function returns a matrix of the relative position index of an attention window
+function get_relative_index(window_size::Tuple)
+    coordinate_h=collect(1:window_size[1]);
+    coordinate_w=collect(1:window_size[2]);#Initialize height and weight coordinates
+    coordinate_pairs=meshgrid(coordinate_h,coordinate_w);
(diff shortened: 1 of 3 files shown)


src/layers/relative_index.jl
```diff
@@ -0,0 +1,32 @@
...
+    relative_coordinate=broadcast.(-,coordinate_flatten_1,coordinate_flatten_2);
+    coordinate_flatten_1=nothing;
+    coordinate_flatten_2=nothing;#RAM management
+    relative_index=Matrix{Any}(undef,window_size[1]*window_size[2],window_size[1]*window_size[2])
+    @threads for i in 1:length(relative_coordinate[1])#use Multi-threading technique to boost performance
+        relative_index[i]=relative_coordinate[1][i];
+        relative_index[i]+=(window_size[1]-1);
+        relative_index[i]*=2*window_size[2]-1;
+        relative_index[i]+=relative_coordinate[2][i]+window_size[2]-1;
+    end
```
ToucheSir, line 30: Adding onto Kyle's point, we generally expect all code that runs in the forward pass of layers to be GPU-compatible as well.

XWHtorrentx: Thank you for your advice.

CarloLucibello:
@XWHtorrentx it is still not clear what you are trying to accomplish with this PR. As @darsnack said, we don't want to provide utility functions in this repo, but just model constructors. Are you going to implement the Swin Transformer or what?

XWHtorrentx:
I am currently working on Swin Transformer. It is almost done and I am testing it. Since I am new here I don’t know that I should pull a model as a whole.

ToucheSir:
I would say open a new PR for the whole model. You can target it against this branch instead of `master` if that would make your life easier too.

darsnack:
Also, we'll help guide your PR so that it is ready to merge. You don't need to worry about it being partially incomplete.

theabhirath (changes requested):
Side note, this PR needs a rebase/reworking on current master as well - a lot of the functionality like `DropPath` is already in (this PR uses a hand-written version), and the API for some of the functions has changed.

theabhirath:
The other thing is the use of NeuralAttentionlib, which last I checked slowed the model down a _lot_ on the CPU. Not sure if there's been any changes there - I will run some benchmarks to check

PR #135, opened 2022-03-15
Refactor of ViT models

theabhirath:
Note: This PR is a logical follow-up to the many refinements made in #125 and thus must be merged only after that one is merged. (Reviewing this after #125 is merged will help too because the changes made in this PR only will be visible more clearly)

This PR refactors the ViT models to make some performance improvements and also to make it easier to write future PRs for attention-based models. First, it takes on [NeuralAttentionlib](https://github.com/chengchingwen/NeuralAttentionlib.jl) as a dep because this allows for writing more general attention operations (which will be very helpful for future models where input dimensions may not be 3 or where there are more complicated hoops to jump through). This has the unexpected side effect of downgrading the Flux version to 0.12.8, however, because the library has a compat bound on NNlib v0.7 and Flux 0.12.9 is the first one to incorporate v0.8. A simple release should solve this, though. (cc @chengchingwen for this)

Overall, I found this version of attention to be significantly faster and inside the ViT model, benchmarking gave me a performance gap of roughly 100 ms for the forward pass. I haven't checked gradient ops due to memory limitations but it seems quite reasonable to assume a performance improvement there as well because NeuralAttentionlib takes care of the gradient for matmul ops, for example.

This also refines the VIT API to use the terminology originally used in the paper for scaling model size (tiny, small, base, large, huge, giant and gigantic).


src/layers/embeddings.jl
```diff
@@ -1,27 +1,34 @@
...
-  h, w, c, n = size(x)
-  hp, wp = h ÷ p.patch_height, w ÷ p.patch_width
-  xpatch = reshape(x, hp, p.patch_height, wp, p.patch_width, c, n)
+  @assert (im_height % patch_height == 0) && (im_width % patch_width == 0)
+  "Image dimensions must be divisible by the patch size."
 
-  return reshape(permutedims(xpatch, (1, 3, 5, 2, 4, 6)), p.patch_height * p.patch_width * c, 
-                 hp * wp, n)
+  return Chain(Conv(patch_size, inchannels => embedplanes; stride = patch_size),
+               flatten ? x -> permutedims(reshape(x, (:, size(x, 3), size(x, 4))), (2, 1, 3))
```
ToucheSir, line 27: This permutedims + reshape routine should be its own function. Internal-only is fine.

src/layers/mlp.jl
```diff
@@ -1,16 +1,44 @@
...
+- `hidden_planes`: Number of dimensions in the intermediate layer.
+- `outplanes`: Number of dimensions in the output - by default it is the same as `inplanes`.
+- `dropout`: Dropout rate.
+- `activation`: Activation function to use.
+"""
+function gated_mlp_block(gate_layer, inplanes::Integer, hidden_planes::Integer,
+												 outplanes::Integer = inplanes; dropout = 0., activation = gelu)
+	@assert hidden_planes % 2 == 0 "`hidden_planes` must be even for gated MLP"
+	return Chain(Dense(inplanes, hidden_planes, activation),
+							 Dropout(dropout),
```
ToucheSir, line 39: Something funky going on with indentation in some files, maybe mixed tabs and spaces?

src/other/mlpmixer.jl
```diff
@@ -1,83 +1,106 @@
...
+- `activation`: the activation function to use in the MLP blocks
 """
-    mlpmixer(imsize::NTuple{2} = (256, 256); inchannels = 3, patch_size = 16, planes = 512, 
-             depth = 12, expansion_factor = 4, dropout = 0., nclasses = 1000, token_mix = 
-             _conv1d, channel_mix = Dense))
+function mixerblock(planes, npatches; mlp_ratio = (0.5, 4.0), mlp_layer = mlp_block, 
+                    dropout = 0., drop_path_rate = 0., activation = gelu)
+  tokenplanes, channelplanes = [Int(r * planes) for r in mlp_ratio]
+  return Chain(SkipConnection(Chain(LayerNorm(planes),
+                                    x -> permutedims(x, (2, 1, 3)),
```
ToucheSir, line 22:
```suggestion
                                    Base.Fix2(permutedims, (2, 1, 3)),
```
Or a separate function since this operation seems to show up in a few places.

test/other.jl
```diff
@@ -2,6 +2,34 @@ using Metalhead, Test
...
+	end
+end
+
+@testset "gMLP" begin
+	@testset for mode in [:small, :base, :large, :huge]
+		@testset for drop_path_rate in [0.0, 0.5, 0.99]
+			m = gMLP(mode; drop_path_rate)
+			@test size(m(rand(Float32, 224, 224, 3, 2))) == (1000, 2)
+			@test_skip gradtest(m, rand(Float32, 224, 224, 3, 2))
+			GC.gc()
```
ToucheSir, line 32: Regarding OOMs, it seems there are a number of models which can accommodate smaller image sizes. I think we can get away with CIFAR-10's 32x32 for those. ImageNet size can be reserved for the smaller variants of each model type if need be, just to make sure the overall architecture is sound.

src/layers/embeddings.jl
theabhirath: This specific combination doesn't show up anywhere but here, though. Still worth it?

src/layers/mlp.jl
theabhirath: Whoops, my bad. I'll fix that

test/other.jl
theabhirath: That messes up the config values for the depth sometimes, since CIFAR is a smaller dataset and sizes may not match - they will for some of the models where the sizes are held constant throughout, thought, I get that. My idea was basically to make sure the config dicts hold up the way they should - we could always shift to a smaller image size wherever we can. But `gradtests` are not even enabled and there are OOMs, so that's definitely a bit of a problem in case those have to be enabled in the future

src/layers/embeddings.jl
ToucheSir: Yes, otherwise the model won't be entirely serializable.

src/layers/attention.jl
```diff
@@ -1,58 +1,48 @@
...
-function (mha::MHAttention)(x)
-  nheads = length(mha.heads.layers)
-  xhead = chunk(x, nheads, dims = 1)
-  return mha.projection(mha.heads(xhead...))
+function (m::MHAttention)(x::AbstractArray{T, 3}) where T
+  B, C, N = size(x)
+  q, k, v = chunk(reshape(m.qkv_layer(x), B ÷ m.nheads, m.nheads, C, 3 * N), 3; dims = 4)
+  scale = convert(T, sqrt(size(q, 1) / m.nheads))
+  attn = m.attn_drop(softmax(NeuralAttentionlib.matmul(q, permutedims(k, (2, 1, 3, 4))) * scale))
+  x = m.projection(reshape(NeuralAttentionlib.matmul(attn, v), (B, C, N)))
```
chengchingwen, line 47: Out of curiosity, why not use `NeuralAttentionlib.multihead_qkv_attention`?

theabhirath: Primarily for demonstration purposes....here of course the library function works perfectly well because it's vanilla attention but for other models there are more things happening and so I wrote it this way to kinda document the internals of self attention for a ViT-like model as more of these models are added. Could always switch - is there a significant performance boost obtained?

chengchingwen: >  is there a significant performance boost obtained?

Probably no, since they work in the same way. But I wonder where does the performance gap (~ 100ms) you mentioned coming from. The functions `NeuralAttentionlib` (currently) provided is to make attention functions composable, and `matmul` should be equivalent to `NNlib.batch_mul` + `reshape`. I'm not sure if the performance gain really comes from `NeuralAttentionlib`

theabhirath: Maybe it's the multiple calls to `Attention` inside `Parallel`? This one simply does MHA directly while the original implementation had `Parallel` with multiple `Attentions` zipped together. The main reason I chose this version was because it also leaves future options open for working with inputs with more than three dimensions if need be (as I understand it `CollapsedDimArray` provides this functionality if necessary)

chengchingwen: > it also leaves future options open for working with inputs with more than three dimensions if need be (as I understand it CollapsedDimArray provides this functionality if necessary

Yes, and actually `NeuralAttentionlib.multihead_qkv_attention` already allows inputs with more dimensions as long as the dimension order is correct.

> while the original implementation had Parallel with multiple Attentions zipped together. 

So this seems to be the main reason for the performance gap. `Parallel` doesn't run in parallel, it sequentially execute each path and aggregate the result, so doing MHA directly would definitely be faster.

theabhirath:
Perhaps I'm getting this wrong, but testing this PR on different versions of Julia has...surprising results.

1.8.0-beta2:
Screenshot 2022-03-17 at 8 22 00 AM

1.7.2:
Screenshot 2022-03-17 at 8 08 39 AM

1.9.0-dev (master):
Screenshot 2022-03-17 at 8 24 18 AM

I can understand the 1.7.2 result, but master should be faster because of all the compiler improvements, right?

ToucheSir:
Compile time regressions are not uncommon on master, but generally speaking they're addressed before an actual release is cut.

src/layers/attention.jl
darsnack: As I recall in the original PR had issues with `batched_mul` across channels, and it also performed the multiplications serially. Does `NeuralAttentionlib` parallelize them?

theabhirath: https://github.com/chengchingwen/NeuralAttentionlib.jl/blob/master/src/matmul/matmul.jl certainly seems to work similar to the NNlib version (but with options for more dimensions), so I think there is parallelism? @chengchingwen could elaborate better though

darsnack: I meant #105 for reference. In that PR, 4D `batched_mul` was originally implemented as a sequential loop over the 3rd dimension (which ended up being faster via `Parallel`). I thought there was already some reshaping going on, but maybe we missed doing it correctly.

Either way, looks like at least the GPU path is parallelized for sure. It's good to consolidate these primitives in NeuralAttentionlib.jl. In #105, part of the problem with 4D `batched_mul` was making assumptions about what the different dimensions mean. In NeuralAttentionlib.jl, it will be much safer to do that.

theabhirath:
Bump?

src/layers/embeddings.jl
```diff
@@ -1,6 +1,9 @@
+_flatify(x) = permutedims(reshape(x, (:, size(x, 3), size(x, 4))), (2, 1, 3))
```
darsnack, line 1: Maybe `flatten_spatial`?

darsnack (changes requested):

src/layers/attention.jl
```diff
@@ -1,58 +1,48 @@
 """
-    Attention(in => out)
-    Attention(qkvlayer)
+    struct MHAttention{P, Q, R}
```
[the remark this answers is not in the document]
theabhirath: Yep, I did want a separate docstring, fixed now!

src/convnets/convnext.jl
```diff
@@ -11,11 +11,11 @@ Creates a single block of ConvNeXt.
 """
 function convnextblock(planes, drop_path_rate = 0., λ = 1f-6)
   layers = SkipConnection(Chain(DepthwiseConv((7, 7), planes => planes; pad = 3), 
-                                x -> permutedims(x, (3, 1, 2, 4)),
+                                permute_dims((3, 1, 2, 4)),
```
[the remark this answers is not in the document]
theabhirath: `swapdims` makes sense, yeah. I didn't really get the `swap_spatial_channels` though - do you mean having convenience functions for interchanging very specific dimensions of the array? If so, would that not make it overly restrictive keeping future models in mind (and also that input sizes may be 4D or 3D)?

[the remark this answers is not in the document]
darsnack: Is it anymore restrictive than what's there right now? Both assume the ordering and number of dimensions. The explicit name would be easier to know what the intent is at a glance.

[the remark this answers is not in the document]
theabhirath: No actually right now `swapdims(perm)` is just `Base.fix2(permutedims, perm)` - so it actually doesn't assume the ordering or the number of dimensions. I _could_ change it to make it more restrictive. Is that something that would make more sense in this particular context?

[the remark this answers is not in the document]
darsnack: I don't mean the definition of `swapdims` which is fine as is. I mean the particular usage in this model (and others) is restricted. But it's okay we can just stick with what we have.

src/utilities.jl
```diff
@@ -38,15 +36,22 @@ Convenient reduction operator for use with `Parallel`.
 """
 cat_channels(xy...) = cat(xy...; dims = 3)
 
+"""
+    swapdims(perm)
+
+Convenience function for permuting the dimensions of an array.
+`perm` is a vector or a tuple of length `ndims(A)` specifying the permutation.
```
darsnack, line 43:
```suggestion
`perm` is a vector or tuple specifying a permutation of the input dimensions.
```

darsnack (approved):

PR #136, opened 2022-03-15
Added UNet architecture (inital commit)

ceferisbarov:
I added the model as-is and updated dependencies and utils. I created a new branch in my fork, but I cannot request to create a new branch in FluxML/Metalhead.jl.

[This](https://github.com/DhairyaLGandhi/UNet.jl) is the original model. I intend to add tests and integrate the model into Metalhead as much as possible.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.7.0-DEV"
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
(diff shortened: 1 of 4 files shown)


lorenzoh:
Agreed, FastAI.jl implementation also works with any backbone architecture by dynamically checking output sizes of each downscaling layer and adding the correct number of upsampling layers with skip connections. Also allows to drop some upscaling layers to get a lower-resolution output. This comes at the cost of not being 100% faithful to the UNet paper; instead it copies the approach from Python's fastai.

ceferisbarov:
Thanks for the feedback. I can try and create a new PR. Any comments on this PR that would be relevant to other models, too?

darsnack:
The overall approach in this PR is good. My only comment would have been to use `SkipConnection`s or `Parallel`, but if you follow FastAI.jl's design, then you should get that. In general, we try not to introduce a new type and `@functor` it unless it makes sense to do so.

ceferisbarov:
I have decided to close this and open a new issue. Otherwise, it will get confusing.

PR #138, opened 2022-03-27
Update README.md

theabhirath:
I've updated the README to reflect some of the newer models and remove some redundancies (specifying the configs for every model but linking the same docstring seemed weird). There is also the fact that the docs still don't seem to be deploying, so the links for the newer models may not work just yet - not sure how that can be resolved


darsnack (approved):
I'll take a look at the docs CI today

PR #140, opened 2022-04-02
Add MobileNet v1

darsnack:
Adds MobileNet v1 for completeness.


src/convnets/mobilenet.jl
```diff
@@ -1,3 +1,105 @@
...
+- `pretrain`: Whether to load the pre-trained weights for ImageNet
+- `nclasses`: The number of output classes
+
+See also [`Metalhead.mobilenetv1`](#).
+"""
+struct MobileNetv1
+  layers
+end
+
+function MobileNetv1(imsize::NTuple{2, Int} = (224, 224), width_mult = 1;
```
ToucheSir, line 88:
```suggestion
function MobileNetv1(imsize::Dims{2} = (224, 224), width_mult = 1;
```

theabhirath: Hmm currently all the models use `NTuple{2, Int}` so maybe we'd have to change that across all occurrences for uniformity

ToucheSir: I wasn't sure if there were any scenarios where `Int` could be something else (some static type, maybe), so I held off from commenting on the earlier model PRs as they left the eltype out. Assuming none of those changes are released, we could always constrain everything to dims and loosen it up if the need arises.

darsnack: There are some which are not intended to be `Dims` but I left those as is. I did change the rest.

darsnack: There are some cases that have `<:Integer`, but hopefully this PR is straightforward enough to get in before v0.7 is released.

darsnack:
There's some mismatch between the "large" tests in #132 and here. Since the tests appear to pass the MobileNet section and error later, I'd like to merge this (assuming no changes requested) and deal with test errors in #132.

src/convnets/mobilenet.jl
```diff
@@ -1,3 +1,99 @@
...
+    MobileNetv1(width_mult = 1; pretrain = false, nclasses = 1000)
+
+Create a MobileNetv1 model with the baseline configuration 
+([reference](https://arxiv.org/abs/1704.04861v1)).
+Set `pretrain` to `true` to load the pretrained weights for ImageNet.
+
+# Arguments
+- `width_mult`: Controls the number of output feature maps in each block
+                (with 1.0 being the default in the paper;
+                 this is usually a value between 0.1 and 1.4)
```
theabhirath, line 73: The original MobileNet had `width_mult` restrictions between 0 and 1, IIRC - with 1.0 being the largest and 0.25, 0.50, 0.75 being possible scaled-down models

darsnack: Thanks, you're right!

darsnack, line 73:
```suggestion
                 this is usually a value between 0 and 1)
```

darsnack:
Bump @ToucheSir?

ToucheSir (approved):

PR #142, opened 2022-04-04
Release 0.7 (round 2)

darsnack:
I rebased #140 on #132 by accident, so #132 is already merged. No problem, it was ready. But this PR is to try and fix some of the memory pressure issues in the tests. Also an opportunity to suggest changes that should have been in #132.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 BSON = "0.3.2"
-Flux = "0.12"
+Flux = "0.13"
 Functors = "0.2"
 MLUtils = "0.1.2, 0.2"
 NNlib = "0.7.34, 0.8"
--- a/docs/tutorials/quickstart.md
+++ b/docs/tutorials/quickstart.md
@@ -18,7 +18,7 @@ Now, we can use this model with Flux like any other model. Below, we train it on
 ```julia
 using Flux: onehotbatch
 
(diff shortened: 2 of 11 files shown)


theabhirath:
Docs CI still seems to be erroring out for some reason 😬 Also, [this error](https://github.com/FluxML/Metalhead.jl/runs/5819603257?check_suite_focus=true#step:6:200) in the CI is one I've encountered occasionally locally as well (particularly when I update Flux and load it for the first time post precompilation). Not sure what the underlying package is, though.

ToucheSir:
Unsure if the warnings have anything to do with the fatal error, but the first is https://github.com/JuliaMath/BFloat16s.jl/pull/37. The second I haven't seen before.

theabhirath:
The second as in the `rpad` one? That's just because Flux 0.12.9 doesn't have MLUtils yet, so the `rpad` definition from Flux and the one from MLUtils (because Metalhead has MLUtils as a dep) are conflicting - should be solved when Flux 0.13 lands

darsnack:
The CI failure might be completely unrelated and due to Publish.jl.

darsnack:
Okay I think the release looks ready. @ToucheSir anything else? @theabhirath I replaced uses of `Diagonal` with the new and improved `Scale`. Are there any replacements possible with Flux 0.13 that I missed?

ToucheSir:
Nope, I think this is good to go.

theabhirath:
Seems perfect!

PR #152, opened 2022-05-01
Try to get gradtests to pass

theabhirath:
Mainly for testing purposes, to see if I can `GC.gc()` enough to make GitHub Actions happy with the memory being used

### DIFF
--- a/test/convnets.jl
+++ b/test/convnets.jl
@@ -6,9 +6,9 @@ PRETRAINED_MODELS = []
 
 @testset "AlexNet" begin
   model = AlexNet()
-  @test size(model(rand(Float32, 256, 256, 3, 1))) == (1000, 1)
+  @test size(model(x_256)) == (1000, 1)
   @test_throws ArgumentError AlexNet(pretrain = true)
-  @test_skip gradtest(model, rand(Float32, 256, 256, 3, 1))
+  @test gradtest(model, x_256)
 end
 
 GC.gc()
@@ -17,13 +17,14 @@ GC.gc()
   @testset "VGG($sz, batchnorm=$bn)" for sz in [11, 13, 16, 19], bn in [true, false]
     m = VGG(sz, batchnorm = bn)
 
-    @test size(m(rand(Float32, 224, 224, 3, 1))) == (1000, 1)
+    @test size(m(x_224)) == (1000, 1)
     if (VGG, sz, bn) in PRETRAINED_MODELS
       @test (VGG(sz, batchnorm = bn, pretrain = true); true)
     else
       @test_throws ArgumentError VGG(sz, batchnorm = bn, pretrain = true)
     end
-    @test_skip gradtest(m, rand(Float32, 224, 224, 3, 1))
+    @test gradtest(m, x_224)
+    GC.gc()
   end
 end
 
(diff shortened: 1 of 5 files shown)


theabhirath:
Well they seem to pass....eventually. The failure on Windows with Julia nightly seems to be persisting across PRs though, is that something that might be an upstream issue?

ToucheSir:
I wonder if sprinkling `yield` or `GC.safepoint` inside the testset loops would help.

For the slow runtime, you could try disabling coverage on all but one config. See https://github.com/FluxML/IRTools.jl/pull/98/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8 for an example of how to do so.

theabhirath:
That's helped timings a little bit, but both the Windows tests are still failing😬

CarloLucibello:
In case we don't manage to fix it, we can conditionally skip: `@test ... skip=Sys.iswindows()`

test/convnets.jl
```diff
@@ -6,177 +6,201 @@ PRETRAINED_MODELS = []
...
+    @test size(m(x_224)) == (1000, 1)
     if (VGG, sz, bn) in PRETRAINED_MODELS
       @test (VGG(sz, batchnorm = bn, pretrain = true); true)
     else
       @test_throws ArgumentError VGG(sz, batchnorm = bn, pretrain = true)
     end
-    @test_skip gradtest(m, rand(Float32, 224, 224, 3, 1))
+    @test gradtest(m, x_224)
+    GC.safepoint()
+GC.gc()
```
DhairyaLGandhi, line 29: indentation

ToucheSir:
What happens if you run the largest configurations first or _only_ test the largest configurations? If the former gets us further or the latter succeeds, then there may be deeper investigation required (not for this PR though).

theabhirath:
> What happens if you run the largest configurations first or _only_ test the largest configurations?

By the largest configurations, do you mean the largest model configurations? (as in `ConvNeXt(:huge)` for example?)

theabhirath:
That...did not help at all

ToucheSir:
That's unfortunate because it means there isn't a quick fix, but fortunate because it eliminates some spooky action at a distance as the culprit. Perhaps Carlo's suggestion is the way to go for now.

theabhirath:
Is there a way to test on a platform with more memory? The Windows failures seem to be related to that (or at least, were a couple of months ago...not so sure now, since the error messages aren't always very descriptive). Might also help with the speed of the tests

theabhirath:
Bump?

test/convnets.jl
```diff
@@ -6,177 +6,201 @@ PRETRAINED_MODELS = []
...
-  @test_skip gradtest(m, rand(Float32, 224, 224, 3, 1))
+  @test gradtest(m, x_224)
 end
 
+GC.safepoint()
 GC.gc()
 
 @testset "DenseNet" begin
-  @testset for sz in [121, 161, 169, 201]
+  @testset for sz in [121, 169, 169, 201]
```
darsnack, line 114: Was this intentional?

darsnack (commented):
We haven't quite figured out how to make this work yet? Or am I missing something.

test/convnets.jl
theabhirath: Oh I was trying something that didn't work out, it should be 161. I'll fix that

theabhirath:
> We haven't quite figured out how to make this work yet? Or am I missing something.

Tests seem to be passing except Windows, where I think the issue is the runners have less memory. Especially because errors like those in #162 tend to sneak by, even if for two OSes having gradtests enabled might be a big help

darsnack:
Agreed, maybe add Carlo's suggestion to skip Windows altogether for grad tests?

theabhirath:
Tests pass 🎉

darsnack (approved):
I guess Windows has issues on master even without grad tests. Maybe disabling all tests completely for now is fine

Issue #117, opened 2022-02-11
Prediction/preprocessing helper


lorenzoh:
Is there interest in a prediction or preprocessing helper that helps people perform on (batches) of data for (Image-Net) pretrained models?

Thinking something like

```julia
function preprocess(img::AbstractArray{U, N}, sz = (224, 224); T = Float32, C = RGB{N0f8} )
    item = DataAugmentation.Image{N}(img)
    tfm = ToEltype(C) |> CenterResizeCrop(sz) |> ToTensor(T)  |> Normalize(IMAGENET_STATS)
    return apply(tfm, item) |> itemdata
end

preprocess(imgs::AbstractVector{<:AbstractArray}) = ... 
```

This would require adding a dep to DataAugmentation.jl, though.

theabhirath:
I'd love to help in whatever way I can....is there a reason that Metalhead doesn't have DataAugmentation built into it though? I've noticed that the Julian way is to keep several lightweight packages and compose them as and when needed but as a torchvision user I often find myself appreciating the fact that everything is in one place 😅 I hope I'm not phrasing this the wrong way, I'm just curious regarding the thought process - I was thinking of writing CV stuff like object detection and semantic segmentation pipelines and was wondering if this would be the right repo to collect everything

lorenzoh:
I'm gathering a lot of common computer vision functionality in FastAI.jl's [`Vision` submodule](https://github.com/FluxML/FastAI.jl/tree/master/src/Vision). FastAI.jl is the batteries-included deep learning package that brings together many of the packages that you end up needing doing deep learning. Unfortunately the documentation for everything but the very high-level cases and interfaces is still lacking; I am planning to tackle this soon, though, to make it easier to contribute to.

For now, if you want to have a preprocessing function like above, try out `ProjectiveTransforms` and `ImagePreprocessing` from FastAI.jl, which end up doing something very similar as the above snippet.

ToucheSir:
One practical reason for splitting packages up currently is import latency. `using Flux`, is already 8-10s, plus another few if you add Metalhead on top of it. If that became negligible then I think something like this would be a no-brainer.

CarloLucibello:
But MLDatasets contains generic datasets, while Metalhead explicitly targets vision. Since vision models are commonly used with preprocessing, I think it would make more sense to have some preprocessing pipeline in here. Or at the very least document how do it using packages from the ecosystem.

CarloLucibello:
[...]

PR #153, opened 2022-05-13
Feature registries for models

theabhirath:
This is a first attempt at using [FeatureRegistries.jl](https://github.com/lorenzoh/FeatureRegistries.jl) to allow the user to view the models available in Metalhead.jl. This is still a work in progress though - most notably, the time taken for `using Metalhead` rises significantly because of the extra code being executed.

It looks pretty though:
Screenshot 2022-05-13 at 4 42 39 PM

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.7.0"
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+FeatureRegistries = "c6aefb4f-3ac3-4095-8805-528476b02c02"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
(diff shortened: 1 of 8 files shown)


darsnack:
Looks great! Seems like this is related to #117, i.e. having a repo separate for "just the models" and then a "vision" repo. The latter can take on heavier deps. Maybe the models should go in MetalheadModels.jl?

theabhirath:
That sounds great! Metalhead can simply call `using MetalheadModels.jl` then. 

One question I have regarding this PR, though, is that the major rise in time for `using Metalhead` seems to be not from `using FeatureRegistries` (which although not negligible, is still only about 200 ms). The rise in time is more around 3-4 seconds, which is mainly because at every `using` call, the parameters of every model are being initiated and summed over. We _could_ theoretically just store the numbers but that doesn't seem so nice. If we continue to give parameter information programmatically, then maybe we need a way to lazy load the numbers only when the function is called - not sure if there is such a method. The other option, of course, is simply to give a different set of information. Would the docstring make more sense here? It would also help the user understand the variants for the model

darsnack:
Yeah a docstring seems more useful than number of parameters. Maybe a description of the configurations too?

PR #154, opened 2022-05-14
Expose `nclasses` option for `DenseNet`

theabhirath:
Previously, the `nclasses` option was not exposed for the `DenseNet` API with the `config` option. 

This PR adds that. Would be good to have a patch release too if possible (only with this commit and not the previous breaking change), since I'm trying to train some models and this will help with that 😅

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Metalhead"
 uuid = "dbeba491-748d-5e0e-a39e-b530a07fa0cc"
-version = "0.7.0"
+version = "0.7.1"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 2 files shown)


darsnack:
Instead of maintaining a separate non-breaking branch, maybe the easiest thing would be to revert the breaking conv + bn changes since they weren't a major impact to TTFG anyways. We can release the other changes along with this and keep the breaking stuff in a separate, new PR.

theabhirath:
> Instead of maintaining a separate non-breaking branch, maybe the easiest thing would be to revert the breaking conv + bn changes since they weren't a major impact to TTFG anyways. We can release the other changes along with this and keep the breaking stuff in a separate, new PR.

I've opened #155 to address this. I will run some more tests when I get time to ascertain just how much returning a `Chain` helps for `conv_bn` but for now, this should help make #151 non-breaking

darsnack:
Can you bump the patch too?

darsnack (approved):

PR #158, opened 2022-05-25
Fix repeat; rewrite multi-head attention layer

### LINKED ISSUE #157: `repeat` is not efficient on GPU
The ViT implementation contains  `repeat`, which will not work well for GPUs (due to scalar indexing)

https://github.com/FluxML/Metalhead.jl/blob/792076f1f86fdf713e2240cc0337c37ffd24fd1c/src/layers/embeddings.jl#L63

An alternative way to simulate repeat with broadcasting

https://github.com/avik-pal/Lux.jl/blob/49b69c7ed21ccde6fc8c07bacaf738de0b6fe497/lib/Boltz/src/vit.jl#L72

theabhirath:
Closes #157. Also borrows from the excellent implementation of `MHAttention` at https://github.com/avik-pal/Lux.jl/pull/29 to remove the NeuralAttentionlib dep, hopefully reducing time for `using Metalhead`

### DIFF
--- a/src/other/mlpmixer.jl
+++ b/src/other/mlpmixer.jl
@@ -17,7 +17,7 @@ Creates a feedforward block for the MLPMixer architecture.
 """
 function mixerblock(planes, npatches; mlp_ratio = (0.5, 4.0), mlp_layer = mlp_block, 
                     dropout = 0., drop_path_rate = 0., activation = gelu)
-  tokenplanes, channelplanes = [Int(r * planes) for r in mlp_ratio]
+  tokenplanes, channelplanes = [Integer(r * planes) for r in mlp_ratio]
   return Chain(SkipConnection(Chain(LayerNorm(planes),
                                     swapdims((2, 1, 3)),
                                     mlp_layer(npatches, tokenplanes; activation, dropout),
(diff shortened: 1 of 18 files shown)


darsnack (changes requested):

darsnack:
It's a bit confusing...fields of structs should be `Int` and functions/constructor arguments should be `Integer`. You want the fields to be a concrete type for performance, you want arguments to be permissive and let the compiler auto-convert when needed.

theabhirath:
> It's a bit confusing...fields of structs should be `Int` and functions/constructor arguments should be `Integer`. You want the fields to be a concrete type for performance, you want arguments to be permissive and let the compiler auto-convert when needed.

Gotcha, that makes sense

src/other/mlpmixer.jl
```diff
@@ -132,7 +132,7 @@ return Chain(SkipConnection(Chain(Flux.Scale(planes),
                                   LayerScale(planes, λ),
                                   DropPath(drop_path_rate)), +),
              SkipConnection(Chain(Flux.Scale(planes),
-                                  mlp_layer(planes, Int(mlp_ratio * planes); dropout, activation),
+                                  mlp_layer(planes, Integer(mlp_ratio * planes); dropout, activation),
```
darsnack, line 135:
```suggestion
                                  mlp_layer(planes, Int(mlp_ratio * planes); dropout, activation),
```

darsnack (approved):
Sorry for the confusion, that was only meant to be a minor suggestion 😅.

theabhirath:
That's quite alright, it helped me understand the rationale better! I've messed up the commits on this one a little bit though, so #159 supersedes this one

PR #159, opened 2022-05-27
Repeat fix 2

### LINKED ISSUE #157: `repeat` is not efficient on GPU

theabhirath:
Supersedes (closes) #158 since I messed up the commits a bit there. Also closes #157.

### DIFF
--- a/src/other/mlpmixer.jl
+++ b/src/other/mlpmixer.jl
@@ -17,7 +17,7 @@ Creates a feedforward block for the MLPMixer architecture.
 """
 function mixerblock(planes, npatches; mlp_ratio = (0.5, 4.0), mlp_layer = mlp_block, 
                     dropout = 0., drop_path_rate = 0., activation = gelu)
-  tokenplanes, channelplanes = [Int(r * planes) for r in mlp_ratio]
+  tokenplanes, channelplanes = [Integer(r * planes) for r in mlp_ratio]
   return Chain(SkipConnection(Chain(LayerNorm(planes),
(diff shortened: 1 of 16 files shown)


src/other/mlpmixer.jl
```diff
@@ -132,7 +132,7 @@ return Chain(SkipConnection(Chain(Flux.Scale(planes),
                                   LayerScale(planes, λ),
                                   DropPath(drop_path_rate)), +),
              SkipConnection(Chain(Flux.Scale(planes),
-                                  mlp_layer(planes, Int(mlp_ratio * planes); dropout, activation),
+                                  mlp_layer(planes, Integer(mlp_ratio * planes); dropout, activation),
```
darsnack, line 135:
```suggestion
                                  mlp_layer(planes, Int(mlp_ratio * planes); dropout, activation),
```

theabhirath: Yikes, seem to have missed that

darsnack (approved):

PR #161, opened 2022-05-29
Demo of SciML style for code

### LINKED ISSUE #160: Code style
Given that the new `sciml` style in JuliaFormatter seems to be working quite well, we could perhaps demo it for this repository? I could try a PR with a `.JuliaFormatter.toml`, but I think the question will be around two major points - indentation for tabs (which is currently 2 but really everywhere else is 4) and spaces around `=` for keyword arguments (which I don't think anyone really agrees on across repositories, so that's just personal preference of picking a style and sticking to it I guess)

theabhirath:
As explained in #160 (will also close #160).. Also adds `.git-blame-ignore-revs` to avoid messing up the blame

### DIFF
--- a/src/convnets/densenet.jl
+++ b/src/convnets/densenet.jl
@@ -10,11 +10,12 @@ Create a Densenet bottleneck layer
                (and scaling factor for inner feature maps; see ref)
 """
 function dense_bottleneck(inplanes, outplanes)
-  inner_channels = 4 * outplanes
-  m = Chain(conv_bn((1, 1), inplanes, inner_channels; bias = false, rev = true)...,
-            conv_bn((3, 3), inner_channels, outplanes; pad = 1, bias = false, rev = true)...)
+    inner_channels = 4 * outplanes
+    m = Chain(conv_bn((1, 1), inplanes, inner_channels; bias = false, rev = true)...,
+              conv_bn((3, 3), inner_channels, outplanes; pad = 1, bias = false,
+                      rev = true)...)
 
-  SkipConnection(m, cat_channels)
+    SkipConnection(m, cat_channels)
 end
 
 """
@@ -27,8 +28,10 @@ Create a DenseNet transition sequence
 - `inplanes`: number of input feature maps
 - `outplanes`: number of output feature maps
 """
-transition(inplanes, outplanes) =
-  Chain(conv_bn((1, 1), inplanes, outplanes; bias = false, rev = true)..., MeanPool((2, 2)))
+function transition(inplanes, outplanes)
+    Chain(conv_bn((1, 1), inplanes, outplanes; bias = false, rev = true)...,
+          MeanPool((2, 2)))
+end
 
 """
     dense_block(inplanes, growth_rates)
@@ -42,8 +45,10 @@ the number of output feature maps by `growth_rates` with each block
 - `growth_rates`: the growth (additive) rates of output feature maps
                   after each block (a vector of `k`s from the ref)
 """
-dense_block(inplanes, growth_rates) = [dense_bottleneck(i, o)
-  for (i, o) in zip(inplanes .+ cumsum([0, growth_rates[1:(end - 1)]...]), growth_rates)]
+function dense_block(inplanes, growth_rates)
+    [dense_bottleneck(i, o)
+     for (i, o) in zip(inplanes .+ cumsum([0, growth_rates[1:(end - 1)]...]), growth_rates)]
+end
 
 """
     densenet(inplanes, growth_rates; reduction = 0.5, nclasses = 1000)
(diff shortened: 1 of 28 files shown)


src/convnets/mobilenet.jl
```diff
@@ -27,37 +27,38 @@ function mobilenetv1(width_mult, config;
...
-  ( true,   64, 1, 1),
-  ( true,  128, 2, 1),
-  ( true,  128, 1, 1),
-  ( true,  256, 2, 1),
-  ( true,  256, 1, 1),
-  ( true,  512, 2, 1),
-  ( true,  512, 1, 5),
-  ( true, 1024, 2, 1),
-  ( true, 1024, 1, 1)
+    #     dw,    c, s, r
```
darsnack, line 51: This slightly misaligned the "columns" for the comment which can be fixed by removing space in front of `dw`.

src/convnets/mobilenet.jl
```diff
@@ -208,71 +213,72 @@ Create a MobileNetv3 model.
...
+                               (3, 6, 80, nothing, hardswish, 2),
+                               (3, 2.5, 80, nothing, hardswish, 1),
+                               (3, 2.3, 80, nothing, hardswish, 1),
+                               (3, 2.3, 80, nothing, hardswish, 1),
+                               (3, 6, 112, 4, hardswish, 1),
+                               (3, 6, 112, 4, hardswish, 1),
+                               (5, 6, 160, 4, hardswish, 2),
+                               (5, 6, 160, 4, hardswish, 1),
+                               (5, 6, 160, 4, hardswish, 1),
+                           ])
```
darsnack, line 277: Another minor point: this might be a case where we override the bot? Can we do that? Not an issue if not.

darsnack (commented):
This is great! I really like the SciML style. Hopefully folks will accept this formatter across FluxML.

src/convnets/mobilenet.jl
theabhirath: So the comment can be fixed if we just keep comma separated characters. The formatter doesn't like it if the spaces aren't uniform and tries to fix it, but I think keep an ordered set of characters even if not lining up with the values below should be good enough>?

darsnack: I meant more that it adds an unnecessary amount of indentation.

theabhirath: Oh. I searched but I don't find an alignment option _exactly_ for this just yet...I'll look some more

darsnack: No it's okay, this is very minor and I was just curious if we can override it. No reason to hold up the change.

darsnack:
This mostly needs a rebase. It might be easier to just open a new PR?

theabhirath:
I _think_ I've covered everything here now

test/convnets.jl
```diff
@@ -5,178 +5,168 @@ using Flux
...
-      @test (VGG(sz, batchnorm = bn, pretrain = true); true)
+        @test (VGG(sz, batchnorm = bn, pretrain = true); true)
     else
-      @test_throws ArgumentError VGG(sz, batchnorm = bn, pretrain = true)
+        @test_throws ArgumentError VGG(sz, batchnorm = bn, pretrain = true)
     end
     @test_skip gradtest(m, rand(Float32, 224, 224, 3, 1))
-  end
-end
+end end
```
ToucheSir, line 28: This looks a little weird, is it a formatter bug?

src/convnets/densenet.jl
```diff
@@ -59,24 +64,24 @@ Create a DenseNet model
...
+        (i != length(growth_rates)) &&
+            push!(layers, transition(outplanes, floor(Int, outplanes * reduction)))
+        inplanes = floor(Int, outplanes * reduction)
+    end
+    push!(layers, BatchNorm(outplanes, relu))
+
+    return Chain(Chain(layers),
+                 Chain(AdaptiveMeanPool((1, 1)),
+                       MLUtils.flatten,
+                       Dense(outplanes, nclasses)))
```
ToucheSir, line 84: This and other calls appear to conflict with the advice "Function calls which cannot fit on a single line within the line limit should be broken up such that the lines containing the opening and closing brackets are indented to the same level while the parameters of the function are indented one level further..." under https://github.com/SciML/SciMLStyle#whitespace

src/convnets/densenet.jl
```diff
@@ -59,24 +64,24 @@ Create a DenseNet model
...
+    layers = []
+    append!(layers, conv_bn((7, 7), 3, inplanes; stride = 2, pad = (3, 3), bias = false))
+    push!(layers, MaxPool((3, 3), stride = 2, pad = (1, 1)))
+
+    outplanes = 0
+    for (i, rates) in enumerate(growth_rates)
+        outplanes = inplanes + sum(rates)
+        append!(layers, dense_block(inplanes, rates))
+        (i != length(growth_rates)) &&
+            push!(layers, transition(outplanes, floor(Int, outplanes * reduction)))
```
ToucheSir, line 76: Unsure if this meets https://github.com/SciML/SciMLStyle#ternary-operator. I suppose `floor(Int, outplanes * reduction)` deserves its own variable either way.

src/convnets/mobilenet.jl
ToucheSir: This alignment also seems contrary to the style guide. Perhaps it's not implemented yet in JuliaFormatter?

src/convnets/densenet.jl
darsnack: I don't quite get that bullet. The guideline and the example below it seem in direct opposition?

test/convnets.jl
theabhirath: Maybe that's just the formatter's way of saving space 😆 But honestly, I have no idea

src/convnets/densenet.jl
ToucheSir: I could've sworn I saw an example like:
```julia
long_function_name(
    long_arg_1_which_requires_a_break,
    long_arg_2_which_requires_a_break,
    ...
)
```
But looking back I cannot find one. Had I looked at other style guides recently this could've been an easy mixup, but I have not. Very confused now...

darsnack: I'm personally a fan of the standard Julia convention which the example and the formatter seem to follow.

PR #169, opened 2022-06-16
Hotfix for ViT on GPU

theabhirath:
Until `chunk` is made GPU-friendly, this should be enough to get ViT up and running on a GPU so that we can cut a patch release (I would like to run some experiments)


ToucheSir (approved):
If I'm not mistaken, this improves type stability as well?

PR #170, opened 2022-06-17
Implementation of `Inceptionv4`, `InceptionResNetv2` and `Xception`

theabhirath:
This PR is my first official contribution as a GSoC student 😁. It does a couple of things:

1. It adds new models (`Inceptionv4` and `InceptionResNetv2`). Internal docs for the layers of these models are left out because they are repetitive and don't really explain much - but if that's a concern I can always fill them in.
2. It deprecates `Inception3` in favour of `Inceptionv3` since the vX notation is being used for models in other libraries (as well as in Metalhead for `MobileNetv3`) and so it makes sense to be explicit about it.

Other things that I happened to fill in because I found them:

1. `cat_channels` is now type-stable as it uses `Val(3)` for `dims` - thanks to @ToucheSir pointing that out in some conversations
2. Catches some formatting issues with #169

Notes:

1. TTFG for the Inception models is off the charts. Particularly, `InceptionResNetv2` is absolutely insane, sometimes taking _minutes_. Subsequent gradients are quite slow as well (in comparison, ViTs, which are heavier models, seem to be faster) and also take a lot of memory. This might be helped by using `Chain(::Vector)` but I haven't used that yet because I'm not sure if it is causing stability issues while training.
2. Closes #131 (supersedes). Should also close #129 since that hasn't seen activity for quite some time.

Edit: added the `Xception` model as well


src/convnets/inception.jl
```diff
@@ -176,21 +178,303 @@ See also [`inception3`](#).
...
+(m::Inceptionv3)(x) = m.layers(x)
+
+backbone(m::Inceptionv3) = m.layers[1]
+classifier(m::Inceptionv3) = m.layers[2]
+
+@deprecate Inception3 Inceptionv3
+
+## Inceptionv4
+
+function mixed_3a()
```
darsnack, line 204: The other inception blocks for v3 allow the number of channels to be configured. It isn't really used right now, but maybe we can still support that here too?

src/convnets/inception.jl
```diff
@@ -176,21 +178,303 @@ See also [`inception3`](#).
...
+    return Parallel(cat_channels, branch1, branch2, branch3, branch4)
+end
+
+function block8(scale = 1.0f0; no_relu = false)
+    branch1 = Chain(conv_bn((1, 1), 2080, 192)...)
+    branch2 = Chain(conv_bn((1, 1), 2080, 192)...,
+                    conv_bn((1, 3), 192, 224; pad = (0, 1))...,
+                    conv_bn((3, 1), 224, 256; pad = (1, 0))...)
+    branch3 = Chain(conv_bn((1, 1), 448, 2080)...)
+    activation = no_relu ? identity : relu
```
darsnack, line 417: Why not directly pass in the activation as a function instead of a keyword?

darsnack (changes requested):
Looks nearly ready right off the bat, nice job! Just a couple initial comments.

src/convnets/inception.jl
theabhirath: This follows the timm design - that too, allows the number of channels to be configure for Inception-v3 but not for v4 or ResNet-v2. My guess is that they're less configurable and so the numbers are too tricky to handle for edge case. But these blocks are never used independently anyways, so I think this should be fine - I'll try and see if they can be added if it still feels like something missing

src/convnets/inception.jl
theabhirath: Yep, was a quirk in the way I wrote it initially before refactoring 🤦🏽 I'll change it up

theabhirath:
CI times are through the roof...we might have to disable gradtests again (or maybe enable it only for select models)

theabhirath:
> CI times are through the roof...we might have to disable gradtests again (or maybe enable it only for select models)

...how on earth are the second CI times so much lesser than the first one?

ToucheSir:
AIUI GHA runners are containerized but not otherwise isolated, so times can vary drastically depending on what else is running on the host.

src/convnets/inception.jl
darsnack: Okay let's skip the customization in this PR

src/Metalhead.jl
```diff
@@ -39,20 +39,21 @@ include("pretrain.jl")
...
        ViT,
-       ConvNeXt, ConvMixer
+       ConvMixer, ConvNeXt
 
 # use Flux._big_show to pretty print large models
-for T in (:AlexNet, :VGG, :ResNet, :GoogLeNet, :Inception3, :SqueezeNet, :DenseNet,
-          :ResNeXt,
-          :MobileNetv1, :MobileNetv2, :MobileNetv3,
-          :MLPMixer, :ResMLP, :gMLP, :ViT, :ConvNeXt, :ConvMixer)
+for T in (:AlexNet, :VGG, :GoogLeNet, :ResNet, :ResNeXt, :Inceptionv3,
```
darsnack: Shouldn't this include the new models?

src/convnets/inception.jl
```diff
@@ -176,21 +178,392 @@ See also [`inception3`](#).
...
+
+@functor Inceptionv3
+
+(m::Inceptionv3)(x) = m.layers(x)
+
+backbone(m::Inceptionv3) = m.layers[1]
+classifier(m::Inceptionv3) = m.layers[2]
+
+@deprecate Inception3 Inceptionv3
+@deprecate inception3 inceptionv3
```
darsnack, line 201: Let's do deprecations in a separate "deprecations.jl" file at the source root. We don't have many, but it makes it easier to check what needs to get dropped on a breaking version.

src/convnets/inception.jl
```diff
@@ -176,21 +178,392 @@ See also [`inception3`](#).
...
+
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
+"""
+struct Inceptionv4
+    layers::Any
+end
+
+function Inceptionv4(; inchannels = 3, dropout = 0.0, nclasses = 1000)
```
darsnack, line 343: This should probably take the `pretrain` keyword for consistency.

theabhirath:
Xception seems to have an even worse gradient benchmark:

```Julia
julia> @benchmark Zygote.gradient(p -> sum($model(p)), $x)
BenchmarkTools.Trial: 1 sample with 1 evaluation.
 Single result which took 9.255 s (78.29% GC) to evaluate,
 with a memory estimate of 10.65 GiB, over 1553316 allocations.
```

src/utilities.jl
```diff
@@ -45,6 +55,14 @@ Equivalent to `permutedims(x, perm)`.
 """
 swapdims(perm) = Base.Fix2(permutedims, perm)
 
+"""
+    applyactivation(activation, x)
+
+Apply an activation function to a given input.
+Equivalent to `activation.(x)`.
+"""
+applyactivation(activation, x) = activation.(x)
```
darsnack, line 64: Is this used somewhere? Did I miss it? I also have trouble seeing how this improves on just writing the broadcast. I would either just do `activation.(x)` or `x -> activation.(x)` in the two scenarios I can think up. Both options are shorter than this.

theabhirath, line 64: Might be a good place to consider upstreaming https://github.com/avik-pal/Lux.jl/blob/aafb2b88455661670ed12055365ba5f573ce4219/src/nnlib.jl#L199 to NNlib and the layer https://github.com/avik-pal/Lux.jl/blob/aafb2b88455661670ed12055365ba5f573ce4219/src/layers/basic.jl#L120 to Flux

src/convnets/inception.jl
```diff
@@ -176,21 +178,403 @@ See also [`inception3`](#).
...
+
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
+"""
+struct InceptionResNetv2
+    layers::Any
+end
+
+function InceptionResNetv2(; inchannels = 3, dropout = 0.0, nclasses = 1000)
```
darsnack, line 470: Same here about `pretrain`

darsnack (changes requested):

src/utilities.jl
theabhirath: This makes the error message (in case there is one) more readable, since with the anonymous function all you can see is `var#XXX`

src/convnets/inception.jl
theabhirath: Will do for the `Inception` line of models, although right now the API for the newer models (like `ViT` or `ConvNeXt`) does not in fact, expose a pretrain option - will have to check that later depending on availability

src/utilities.jl
darsnack: Wouldn't `x -> applyactivation(activation, x)` have the same issue?

theabhirath: Right now it's `Base.Fix1`, which I'm _hoping_ will error slightly more clearly. But see https://github.com/FluxML/Metalhead.jl/pull/170#discussion_r900694370

darsnack: I'm not a fan of needing to wrap activations like that in a layer. NNlib already provides a way for layers to use a fast activation implementation e.g. `tanh_fast`. I don't see why GPU activations can't hook into the same mechanism.

src/convnets/inception.jl
darsnack: Hmm those probably should to but we can fix that later. The `loadpretrain!` mechanism already does the error handling for you. So you can unconditionally do the loading when `pretrain` is true.

src/utilities.jl
theabhirath: Mostly this is because of error messages. With anonymous functions, the error is unclear. While part of Flux's advantage is writing it that way just works, it also means that for newer uses when stuff errors out it is kinda hard to track down

darsnack: The easiest is to just make the default activations broadcast on arrays (which I think NNlib already does for some of them). Better might be adding error hinting to them (i.e. `relu6(::AbstractArray)` -> "MethodError: maybe you intended to broadcast relu6 over an array?"

theabhirath: Hmm, true. This might need to be taken up in NNlib then. I'll do away with `applyactivation` for now and make it a plain broadcast

darsnack: I totally agree but I think Julia provides us better options than creating essentially a domain specific language for function calling (which is what this solution is). Perhaps if you think this is nice but not necessary for the PR, we can open an issue on Flux for more discussion? I agree that having a nice error message across all activations would be good.

Also, it seems NNlibCUDA already dispatches to the faster activations without explicitly needing anything from the user. So the performance detail is moot.

theabhirath: That makes sense. I will open up an issue

theabhirath:
This _should_ be ready now, although I'm not sure what exactly I am to make of the gradient times. I suppose they shouldn't block this PR, but is there something that could be a possible contributor that I _can_ change?

src/utilities.jl
theabhirath: https://github.com/FluxML/NNlib.jl/issues/422

theabhirath:
The memory problems return 😬

src/convnets/mobilenet.jl
```diff
@@ -36,7 +36,7 @@ function mobilenetv1(width_mult, config;
         for _ in 1:nrepeats
             layer = dw ?
                     depthwise_sep_conv_bn((3, 3), inchannels, outch, activation;
-                                          stride = stride, pad = 1) :
+                                          stride = stride, pad = 1, bias = false) :
```
darsnack, line 39: Was this change intended? Is this a bug you caught?

src/layers/conv.jl
```diff
@@ -1,8 +1,8 @@
 """
     conv_bn(kernelsize, inplanes, outplanes, activation = relu;
-            rev = false, preact = true,
-            stride = 1, pad = 0, dilation = 1, groups = 1, [bias, weight, init],
-            initβ = Flux.zeros32, initγ = Flux.ones32, ϵ = 1f-5, momentum = 1f-1)
+                 rev = false, preact = false, use_bn = true,
+                 initβ = Flux.zeros32, initγ = Flux.ones32, ϵ = 1.0f-5, momentum = 1.0f-1,
+                 kwargs...)
```
darsnack, line 5: For a doc string this should list the optional keywords like before.

src/layers/conv.jl
```diff
@@ -71,6 +77,8 @@ See Fig. 3 in [reference](https://arxiv.org/abs/1704.04861v1).
   - `outplanes`: number of output feature maps
   - `activation`: the activation function for the final layer
   - `rev`: set to `true` to place the batch norm before the convolution
+  - `use_bn1`: set to `true` to use a batch norm after the depthwise convolution
+  - `use_bn2`: set to `true` to use a batch norm after the pointwise convolution
```
darsnack, line 81: Maybe combine these into a single keyword with 2-elements like `(true, true)`?

darsnack (changes requested):

darsnack:
Not sure there's much we can do beyond what's already been tried in Metalhead. We just need to do something about `applychain` in Flux/Zygote.

src/convnets/mobilenet.jl
theabhirath: Yep, this is intended. MobileNetv1 still needs to be fixed a little bit, but I will leave that for a separate PR

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+            inc = i == 1 ? inchannels : outchannels
+            outc = outchannels
+        else
+            inc = inchannels
+            outc = i == nrepeats ? outchannels : inchannels
+        end
+        push!(layers, x -> relu.(x))
+        append!(layers,
+                depthwise_sep_conv_bn((3, 3), inc, outc; pad = 1, bias = false,
+                                      use_bn = (false, false)))
```
darsnack, line 527: Is there a use case of `(true, false)` or `(false, true)`? If not, I was thinking it is kinda silly to use `conv_bn` with the keyword `use_bn = false` cause that's just `Conv`. It makes sense for consistency with `depthwise_sep_conv_bn`. But it might be cleaner to just introduce `depthwise_sep_conv` then instead of more keywords in multiple places?

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+([reference](https://arxiv.org/abs/1610.02357))
+
+# Arguments
+
+  - inchannels: number of input channels.
+  - outchannels: number of output channels.
+  - nrepeats: number of repeats of depthwise separable convolution layers.
+  - stride: stride by which to downsample the input.
+  - start_with_relu: if true, start the block with a ReLU activation.
+  - grow_first: if true, increase the number of channels at the first convolution.
```
darsnack, line 504:
```suggestion
  - `inchannels`: number of input channels.
  - `outchannels`: number of output channels.
  - `nrepeats`: number of repeats of depthwise separable convolution layers.
  - `stride`: stride by which to downsample the input.
  - `start_with_relu`: if true, start the block with a ReLU activation.
  - `grow_first`: if true, increase the number of channels at the first convolution.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+"""
+function xception_block(inchannels, outchannels, nrepeats; stride = 1,
+                        start_with_relu = true,
+                        grow_first = true)
```
darsnack, line 508:
```suggestion
                        grow_at_start = true)
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+    inceptionresnetv2(; inchannels = 3, dropout = 0.0, nclasses = 1000)
+
+Creates an InceptionResNetv2 model.
+([reference](https://arxiv.org/abs/1602.07261))
+
+# Arguments
+
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
```
darsnack, line 431:
```suggestion
  - `inchannels`: number of input channels.
  - `dropout`: rate of dropout in classifier head.
  - `nclasses`: the number of output classes.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+
+Creates an Inceptionv4 model.
+([reference](https://arxiv.org/abs/1602.07261))
+
+# Arguments
+
+  - `pretrain`: set to `true` to load the pre-trained weights for ImageNet
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
```
darsnack, line 331:
```suggestion
  - `inchannels`: number of input channels.
  - `dropout`: rate of dropout in classifier head.
  - `nclasses`: the number of output classes.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+    inceptionv4(; inchannels = 3, dropout = 0.0, nclasses = 1000)
+
+Create an Inceptionv4 model.
+([reference](https://arxiv.org/abs/1602.07261))
+
+# Arguments
+
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
```
darsnack, line 291:
```suggestion
  - `inchannels`: number of input channels.
  - `dropout`: rate of dropout in classifier head.
  - `nclasses`: the number of output classes.
```

darsnack (changes requested):
Only some minor doc stuff

src/convnets/inception.jl
theabhirath: I can't say I see something _yet_ but I think this can be kept for some time...with all the churn over the next two months it might be handy. If not, I'll simplify it later

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+    xception(; inchannels = 3, dropout = 0.0, nclasses = 1000)
+
+Creates an Xception model.
+([reference](https://arxiv.org/abs/1610.02357))
+
+# Arguments
+
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
```
darsnack, line 545:
```suggestion
  - `inchannels`: number of input channels.
  - `dropout`: rate of dropout in classifier head.
  - `nclasses`: the number of output classes.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+
+Creates an Xception model.
+([reference](https://arxiv.org/abs/1610.02357))
+
+# Arguments
+
+  - pretrain: set to `true` to load the pre-trained weights for ImageNet.
+  - inchannels: number of input channels.
+  - dropout: rate of dropout in classifier head.
+  - nclasses: the number of output classes.
```
darsnack, line 576:
```suggestion
  - `pretrain`: set to `true` to load the pre-trained weights for ImageNet.
  - `inchannels`: number of input channels.
  - `dropout`: rate of dropout in classifier head.
  - `nclasses`: the number of output classes.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+
+!!! warning
+    
+        `Xception` does not currently support pretrained weights.
```
darsnack, line 580:
```suggestion
    `Xception` does not currently support pretrained weights.
```

src/convnets/inception.jl
```diff
@@ -176,21 +174,419 @@ See also [`inception3`](#).
...
+# Arguments
+
+  - `pretrain`: set to `true` to load the pre-trained weights for ImageNet
+  - `inchannels`: number of input channels.
+  - `dropout`: rate of dropout in classifier head.
+  - `nclasses`: the number of output classes.
+
+!!! warning
+    
+        `Inceptionv4`` does not currently support pretrained weights.
```
darsnack, line 335:
```suggestion
    `Inceptionv4`` does not currently support pretrained weights.
```

darsnack (changes requested):

theabhirath:
Ah sorry, I keep missing these minor doc tweaks 🤦🏽 I'll be a bit more thorough next time

darsnack (approved):
No problem! Thanks and nice job!

PR #164, opened 2022-06-09
Add pre-trained models from HuggingFace and add accuracy testing

darsnack:
This PR builds on the great work by @Alexander-Barth in https://github.com/FluxML/MetalheadWeights/pull/2 to add pre-trained weights for VGG and ResNet from torchvision.

We ran into issues with Github LFS limits on MetalheadWeights. Instead, we are using HuggingFace to host our models now which is really ideal for our use case.

This PR adds:
- [x] Pre-trained models for VGG (w/o BN) + ResNet
- [x] Updated quickstart guide
- [x] Updated contributor docs for how to others can add weights via HF
- [x] Refactored tests to check the accuracy of pre-trained models (by just checking against a single reference image from ImageNet)

Close #86, closes #72, and closes #143.

PR #171, opened 2022-06-19
Implement EfficientNet

### LINKED ISSUE #106: Add EfficientNet with pretrained weights
@pxl-th has implemented EfficientNet models in his repo https://github.com/pxl-th/EfficientNet.jl and also implemented the infrastructure to load the weights from pytorch's checkpoints. If the author agrees, we could copy or move the code here and possibly extend the weight loading to other models.

darsnack:
Follow-up from #106 (closes #106).

### DIFF
--- a/src/convnets/efficientnet.jl
+++ b/src/convnets/efficientnet.jl
@@ -13,7 +13,7 @@ Create an EfficientNet model ([reference](https://arxiv.org/abs/1905.11946v5)).
     - `k`: kernel size
     - `s`: kernel stride
     - `e`: expansion ratio
-    - `i`: block input channels
+    - `i`: block input channels (will be scaled by global width scaling)
     - `o`: block output channels (will be scaled by global width scaling)
 - `inchannels`: number of input channels
 - `nclasses`: number of output classes
@@ -110,7 +110,7 @@ See also [`efficientnet`](#).
     - `k`: kernel size
     - `s`: kernel stride
     - `e`: expansion ratio
-    - `i`: block input channels
+    - `i`: block input channels (will be scaled by global width scaling)
     - `o`: block output channels (will be scaled by global width scaling)
 - `inchannels`: number of input channels
 - `nclasses`: number of output classes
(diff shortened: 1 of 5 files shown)


src/convnets/efficientnet.jl
```diff
@@ -0,0 +1,155 @@
...
+- `inchannels`: number of input channels
+- `nclasses`: number of output classes
+- `max_width`: maximum number of output channels before the fully connected
+               classification blocks
+"""
+function efficientnet(scalings, block_config;
+                      inchannels = 3, nclasses = 1000, max_width = 1280)
+    wscale, dscale = scalings
+    out_channels = _round_channels(32, 8)
+    stem = Chain(Conv((3, 3), inchannels => out_channels;
```
theabhirath, line 27: Isn't this just `conv_bn`?

src/convnets/efficientnet.jl
```diff
@@ -0,0 +1,155 @@
...
+    - `s`: kernel stride
+    - `e`: expansion ratio
+    - `i`: block input channels
+    - `o`: block output channels (will be scaled by global width scaling)
+- `inchannels`: number of input channels
+- `nclasses`: number of output classes
+- `max_width`: maximum number of output channels before the fully connected
+               classification blocks
+"""
+function EfficientNet(scalings, block_config;
```
theabhirath, line 119: Isn't exported yet, I reckon

theabhirath:
Is there also a way we can port the weights functionality from #113? Would be a good complement to #164

darsnack:
Probably using the same technique as #164. I think any model that is plain convolutions, batch norm, and fully connected can be done in this way. There a gist linked on the HF repo model cards if you want to tackle adding this for other models in the repo.

darsnack:
Test errors are happening later for the same memory issues as always. I've already limited the tests to the smallest subset that seems necessary.

theabhirath:
Seems all good, but I'm worried that EfficientNets, which are the least heavy of the models, are causing this....or maybe it's a cumulative effect from the other models (Is `GC.gc()` perfect?)

theabhirath (approved):

darsnack:
The tests get passed EfficientNet and error around ConvNeXt. Maybe ConvNeXt is right on the threshold for the CI?

theabhirath:
Seems like it was the MLPMixers after all, nice!

theabhirath:
Could maybe try not testing for the `ConvNeXt` and `ConvMixer` large models, but I'm not sure if those are the only pain points for memory here - `Xception` and `InceptionResNetv2` were pretty bad as well

PR #182, opened 2022-07-09
Initial switch to ReTest.jl

darsnack:
This switches the testing framework to ReTest.jl (already being used for FastAI.jl CI). We should be able to define jobs that filter the test suite so that different runners only run a portion of the test suite (the downside being that we will have _many_ jobs). Unlike a solution based on clever `include`s, this should also allow developers to run an even finer subset of the tests locally (e.g. when adding a new model).


theabhirath:
Nightly tests should be failing but it shows successful? 🤔

darsnack:
I allowed failures on nightly in GHA.

ToucheSir:
Should we get rid of nightly MacOS CI while we're at it and save a few cycles?

theabhirath:
Could disable all nightly CIs and enable Windows instead, which is disabled as of now. And if that's still too much, could always disable Linux since macOS should catch any *nix OS-specific errors and keep macOS and Windows

ToucheSir:
I'd rather avoid testing on mac wherever possible because those runners are harder to come by. One idea could be to split out a dedicated suite for all the model bits, utilities etc. and only test that on Windows/Mac/maybe nightly. Then the E2E tests can be the sole domain of latest Julia on Linux/maybe nightly.

darsnack:
I decided to just disable the macOS runners for now. On this launch, it looks like all the sub-suites launched in parallel, so hopefully that makes typical turn around time much faster.

Separating the "utilities" from the end-to-end tests is a good idea. That will let us have best-effort testing on other systems and make reverse CI faster too. But I am going to leave that for another PR since it requires actually writing the "utility" tests.

darsnack:
One last thing is getting coverage to work right. Each entry in the matrix appears to generate the coverage data correctly, but we probably want to run the processing separately after the entire matrix is complete. I couldn't figure out how to share files across jobs.

theabhirath:
I _think_ ConvMixer and ConvNeXt together should be okay (`:xlarge` on ConvNeXt might cause a trip-up, though - if it does, disabling it might not be _that_ bad). Also might be nice to just make "Others" explicit as MLPMixer variants. There's also some weird regex around ResNet, not sure if that's necessary. Would be nice to get this in as quickly as possible so that I can actually see CI pass on my PRs (small feelings of satisfaction 😄 )

Not a priority, but is there a way to also trigger only specific CI tests on GHA using ReTest? Like if I've not touched the ViTs, is there a way to tell GHA in a commit to not trigger those tests? This could be file based maybe - any changes in `Layers` should cause all tests to trigger, for example, but modifying a ResNet and getting the entire suite to run seemed overkill

ToucheSir (approved):
Looks like this was a very successful experiment: 30mins max on CI and all greens!

darsnack:
> Not a priority, but is there a way to also trigger only specific CI tests on GHA using ReTest? Like if I've not touched the ViTs, is there a way to tell GHA in a commit to not trigger those tests? This could be file based maybe - any changes in Layers should cause all tests to trigger, for example, but modifying a ResNet and getting the entire suite to run seemed overkill.

Yeah I use [this action here](https://github.com/UW-PHARM/BitSAD.jl/blob/68f64376960280b4f6f011cd2bb6972e9c21a481/.github/workflows/Artifacts.yml#L16-L20) in another package. We might be able to do the same thing here, but we can save it for another PR.

darsnack:
How is Inception still running 😕

PR #174, opened 2022-06-21
Overhaul of ResNet API

theabhirath:
This PR completely re-writes the current ResNet API to make it more powerful, more extensible and to reduce code duplication.

## Why this PR?

While making ResNet more fully-featured, this PR will also:
1. Add support for [DropBlock](https://arxiv.org/abs/1810.12890) to ResNet, a type of regularisation used in place of Dropout in some networks
2. Add support for an optional deeper stem front-up
3. Add support for attention layers
4. Add support for multiple pooling options in the classifier head
5. Allow for more ResNet block variants (such as those from the [Bag of Tricks](https://arxiv.org/pdf/1812.01187) paper)

## Things to do

- [x] Abstract out pooling, classifier heads to provide more options
- [x] Add attention layer
- [x] Specific constructors for the ResNet paper models
- [x] Documentation, Documentation and more documentation
  - [x] `DropBlock` and its behaviour
  - [x] `DropPath` behaviour in detail (permissible values, calculations)
  - [x] Higher level `ResNet` interface vs lower level `resnet` interface
- [x] Rewrite ResNeXt to use the new `resnet` API
- [x] Write more tests
- [x] Benchmark to make sure there are no regressions
  - [x] Forward pass
  - [x] Backward pass
  - [x] TTFG
- [ ] Figure out a way to port pre-trained weights

## Other PRs to land before this one

- [x] activation functions in `Chain`s directly without broadcasting (https://github.com/FluxML/NNlib.jl/pull/423). 
     - [x] Needs a patch release.
- [x] `rand_like` and `randn_like` in MLUtils (https://github.com/JuliaML/MLUtils.jl/pull/101)
    - [x] Needs a patch release

## Miscellaneous fixes

- Adds a type argument to `densenet` for `nblocks` to avoid hitting integer edge cases

PR #184, opened 2022-07-24
Add more options to `ViT`

theabhirath:
This modifies the current ViT API to add more options - notably, there is now a optional `prenorm/postnorm` toggle. There's also an option to make class tokens disappear completely, and also to allow class tokens to be before the positional embedding as in DeIT-III. This makes some other cleanup changes as well. The API is more congested for now but I thought I'd get this in before I start working on the other ViTs - maybe there's some potential for extracting common stuff out there.
Needs #174 to land before this makes sense. Also documentation is pending


zsz00:
what's status now ?

theabhirath:
I've had less time to work on this in the recent past, but I'm going to try and push some of these refactors through in the next few months. However, there's some work that's also happening on the Attention implementations around the Flux ecosystem - I suspect any reforms to the ViT will wait on that work to land

Issue #186, opened 2022-07-26
MobileNetv1 doesn't match the implementation in the paper


theabhirath:
Currently MobileNetv1 has an extra `Dense` layer [here](https://github.com/FluxML/Metalhead.jl/blob/48d1401153355e399df0b22470602da0d8016aff/src/convnets/mobilenet.jl#L50), which means that it differs from the implementation described in the paper and other common libraries.

PR #189, opened 2022-07-28
Removed extra mobilenetv1 fc layer, fcsize param

cluffa:
#186  
My first PR! Thanks @theabhirath for a good first issue. I found that the best way to fix was to also remove the `fcsize` parameter. It's not used by the non-configurable function, so I'm assuming it's not used much.  

It's also an option to set `fcsize = nothing` and add extra layer if `fcsize !== nothing`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Metalhead"
 uuid = "dbeba491-748d-5e0e-a39e-b530a07fa0cc"
-version = "0.7.3"
+version = "0.8.0-DEV"
 
 [deps]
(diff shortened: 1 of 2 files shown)


theabhirath (approved):
Thanks for the PR, seems good to me! Maybe @darsnack can comment on keeping the extra FC layer optionally?

darsnack (changes requested):
No, your proposed changes look great! I don't think we need to maintain the ability to have an intermediate layer.

Can you update the version in the Project.toml to `0.8.0-DEV` so that we don't forget that there is now a breaking change for release?

Issue #176, opened 2022-06-23
[Feature request] Add `inchannels` and `nclasses` as kwarg for all constructors


ablaom:
I'm looking at Metalhead integration in MLJFlux. To do this well, I'm looking for some uniformity in the Metalhead.jl API that seems to be lacking. In particular, it would help if `nclasses` and `inchannels` were supported kwargs for all Metalhead constructors (`VGG`, and so forth).   Perhaps if there are models that really can't support them, an informative error could be thrown?

For example, I think the following ought to just work but doesn't:

```julia
VGG(inchannels=1)
```

The same comment applies to `nclasses` but I haven't checked that one to see if there are gaps.

Also helpful would be a trait to label those constructors that build *fixed image size* models. I think there's at least one.

theabhirath:
`nclasses` is supported by every current Metalhead model, I believe. `inchannels` isn't but I have been working on trying to unify the API a little bit, and so it can be easily done as part of that. 

> Also helpful would be a trait to label those constructors that build _fixed image size_ models. I think there's at least one.

This might be a little complicated. Some models explicitly allow for an `imsize` parameter to be passed in. And while others don't, that does not necessarily mean they don't work. The Inception series is a good example - while being designed for 299x299 RGB images, they also sometimes end up working with lower resolutions. I'll see if there's a way I can ascertain this for sure

ablaom:
Thanks for the lightning feedback, @theabhirath. 

> nclasses is supported by every current Metalhead model, I believe. inchannels isn't but I have been working on trying to unify the API a little bit (see https://github.com/FluxML/Metalhead.jl/pull/174 for ResNet, for example), and so it can be easily done as part of that.

👍🏾 

> This might be a little complicated. 

In my use-case an `imsize` is always available. Perhaps all models also accept `imsize`, but throw error/warning if that choice of `imsize` is definitely/possilby going to lead to unexpected behaviour.

darsnack:
I think we can support image size, input channels, and output classes as consistent arguments to the highest level constructors (i.e. the ones that have the `pretrain` keyword). I think this should be good for MLJ, since you plan to support standard architectures with tweaks on the input/output side only?

We can probably have a non-breaking release that adds the interface for this, but not all the models might fully support all the knobs. Once @theabhirath's work lands though, I think we'll be able to support it for every CNN-like architecture.

ablaom:
[...]

PR #190, opened 2022-08-01
Expose a uniform API at the highest level for models

theabhirath:
This PR does a bunch of stuff:

1. Proposes to refine the API at the highest level of models to only expose options that make sense for the model to load pretrained weights (this means removing configurations like drop rates). 
2. Completes the migration towards having `inchannels` and `nclasses` as uniform options for all models, helping along #176 - only `imsize` is left now. 
3. For the purpose of general code cleanliness, defines a _ton_ of type annotations that weren't there before - these will help along with the documentation (once refactored) to help the users figure out more cryptic errors that may occur when dealing with lower level model APIs.  
4. Does away with default model configurations at the highest level. I can restore this, but this behaviour is rather ambiguous - what's to make a ResNet 50 be more suited for a default configuration than a ResNet 18, or a 101?
5. Throws in a small refactor of `invertedresidual` as prep for an EfficientNetv2 PR that will be landing shortly.
6. Uses `create_classifier` in most cases where it can be for brevity
7. Adds compat entries for packages introduced in #174 (closes #191, closes #192, closes #193, closes #194). Also closes #180 because why not


theabhirath:
Pending CI, this should be good to go!

darsnack (approved):

PR #198, opened 2022-08-11
Implementation of EfficientNetv2 and MNASNet

theabhirath:
This is an implementation of [EfficientNetv2](https://arxiv.org/abs/2104.00298) and [MNASNet](https://arxiv.org/abs/1807.11626). There's clearly some code duplication between the structures of the `efficientnet` function and the `mobilenetvX` functions, but figuring out how to unify that API and its design is perhaps best left for another PR.

## TODO
- [x] Figure out a way to unify the EfficientNet and EfficientNetv2 lower level API into a single `efficientnet` function
- [x] More elaborate docs for `mbconv` and `fused_mbconv`

P.S. memory issues mean that the tests are more fragmented than they ought to be, not sure how we can go about addressing those in the short/medium term


src/convnets/mobilenets/mobilenetv3.jl
```diff
@@ -19,38 +19,49 @@ Create a MobileNetv3 model.
...
     inplanes = _round_channels(16 * width_mult, 8)
     layers = []
     append!(layers,
             conv_norm((3, 3), inchannels, inplanes, hardswish; stride = 2, pad = 1))
     explanes = 0
+    nstages = length(configs)
+    reduced_divider = 1
     # building inverted residual blocks
-    for (k, t, c, reduction, activation, stride) in configs
+    for (i, (k, t, c, reduction, activation, stride)) in enumerate(configs)
```
theabhirath, line 40: While I work on refactoring the EfficientNet, this part is something that's been annoying me so I decided to put it up. `torchvision` has a cool feature from other papers (see https://pytorch.org/blog/torchvision-mobilenet-v3-implementation/ for a proper explanation) where they use dilations and a reduced tail (i.e. last three blocks) for some engineering gains. My code works but looks terribly ugly - any idea how to make this look prettier?

ToucheSir: Pull it out into a utility function?

theabhirath: That's definitely a possibility, but I don't like the fact that this changes the implementation details for MobileNetv3 so fundamentally that unifying the code for the three MobileNets now becomes a nightmare. `torchvision` uses class variables, so they are able to get away with writing [this](https://github.com/pytorch/vision/blob/96aa3d928c6faaf39defd846966421679244412d/torchvision/models/mobilenetv3.py#L239-L285) for example - any creative ways to do something similar in Julia?

ToucheSir: Isn't that a utility function? :P Perhaps I'm missing something.

theabhirath: > Isn't that a utility function? :P Perhaps I'm missing something.

It is, but `torchvision` writes the code for the three MobileNets differently, and I wanted to avoid that if possible in Julia. But what I more meant was that the tail dilation and reduced tail are built into the config dict because all of those variables are in the same function scope

ToucheSir: Is the extent of the differing logic here the conditional below? If so, perhaps it could be its own function, parameterized on `i` or some boolean to indicate the calculation is being done for a tail layer. This function has a branch for whether you want the dilations and/or reduced dimensions at the tail. Won't be the prettiest thing, but it doesn't have to be either.

theabhirath:
Side note - using `basic_conv_bn` for the Inception models seems to have fixed their gradient times somehow, which are now _much_ more manageable

src/convnets/efficientnets/core.jl
```diff
@@ -0,0 +1,66 @@
...
+    end
+    return get_layers, nrepeats
+end
+
+function mbconv_stack_builder(block_configs::AbstractVector{<:Tuple},
+                              residual_fns::AbstractVector; inplanes::Integer,
+                              scalings::NTuple{2, Real} = (1, 1),
+                              norm_layer = BatchNorm)
+    bxs = [residual_fn(block_configs, inplanes, stage_idx; scalings, norm_layer)
+           for (stage_idx, residual_fn) in enumerate(residual_fns)]
```
darsnack, line 42: Can we translate between `mbconv` and `mbconv_builder` here? This would make it so that `efficientnet` accepts a vector of `mbconv`/`fused_mbconv` directly. Then the mid-level API has no mention of builders at all.
```suggestion
    _get_builder(::typeof(mbconv)) = mbconv_builder
    _get_builder(::typeof(fused_mbconv)) = fused_mbconv_builder
    bxs = [_get_builder(residual_fn)(block_configs, inplanes, stage_idx;
                                     scalings, norm_layer)
           for (stage_idx, residual_fn) in enumerate(residual_fns)]
```

src/convnets/inceptions/inceptionv3.jl
```diff
@@ -33,10 +33,10 @@ Create an Inception-v3 style-B module
   - `inplanes`: number of input feature maps
 """
 function inceptionv3_b(inplanes)
-    branch3x3_1 = Chain(conv_norm((3, 3), inplanes, 384; stride = 2))
-    branch3x3_2 = Chain(conv_norm((1, 1), inplanes, 64)...,
-                        conv_norm((3, 3), 64, 96; pad = 1)...,
-                        conv_norm((3, 3), 96, 96; stride = 2)...)
+    branch3x3_1 = Chain(basic_conv_bn((3, 3), inplanes, 384; stride = 2)...)
```
darsnack, line 36: If the timing difference is inference, could this `...` have made the difference? There are a couple places where this occurs, and mixing `Chain(::Vector)` and `Chain(::Tuple)` sounds like a recipe of problems.

src/convnets/mobilenets/mobilenetv3.jl
```diff
@@ -20,45 +16,57 @@ Create a MobileNetv3 model.
...
+    (with 1 being the default in the paper; this is usually a value between 0.1 and 1.4.)
   - `max_width`: The maximum number of feature maps in any layer of the network
+  - `dropout_rate`: The dropout rate to use in the classifier head. Set to `nothing` to disable.
+  - `inchannels`: The number of input channels.
   - `nclasses`: the number of output classes
 """
-function mobilenetv3(width_mult::Real, configs::AbstractVector{<:Tuple};
-                     max_width::Integer = 1024, inchannels::Integer = 3,
-                     nclasses::Integer = 1000)
+function mobilenetv3(configs::AbstractVector{<:Tuple}; width_mult::Real = 1,
```
darsnack, line 27: This is going to get updated now that the `SkipConnection`s are handled in `resnet` right?

darsnack (changes requested):
Overall looks good; the design is ready, I think.

src/convnets/mobilenets/mobilenetv3.jl
theabhirath: This went away in the recent refactor, but keeping this conversation unresolved as a reminder that maybe some day we'll find a way to bake this into the configuration dict

src/layers/conv.jl
```diff
@@ -25,127 +26,59 @@ Create a convolution + batch normalization pair with activation.
...
             activations = (conv = activation, bn = identity)
         end
     end
-    layers = [Conv(kernel_size, inplanes => outplanes, activations.conv; kwargs...),
-        norm_layer(bnplanes, activations.bn)]
+    # handle momentum for BatchNorm
+    if !isnothing(momentum)
+        @assert norm_layer==BatchNorm "`momentum` is only supported for `BatchNorm`"
+        norm_layer = (args...; kwargs...) -> BatchNorm(args...; momentum, kwargs...)
+    end
```
darsnack, line 66: Originally, we had the `BatchNorm` arguments specifically listed. Then with `norm_layer`, we removed them. The reason I liked that change is that it avoids `kwargs...` passing. Meaning that if I want a `BatchNorm` with a specific momentum, then the caller should be doing `conv_norm(...; norm_layer = (args...; kwargs...) -> BatchNorm(args...; momentum = MY_NUM, kwargs...)`. Then you can also remove all the `momentum` / `nothing` handling from `layers/mbconv.jl`.

Right now, we are just special casing one argument, so it seems okay. But this type of design scales poorly as we need more arguments. The closure based one via `norm_layer` scales well to arbitrary norm layers and their arguments. Instead, `mnasnet` should specify `norm_layer` as above.

test/convnets.jl
```diff
@@ -264,39 +278,54 @@ end
...
-    _gc()
-    @testset "MobileNetv2" begin
-        m = MobileNetv2()
-        @test size(m(x_224)) == (1000, 1)
-        if MobileNetv2 in PRETRAINED_MODELS
-            @test acctest(MobileNetv2(pretrain = true))
-        else
-            @test_throws ArgumentError MobileNetv2(pretrain = true)
+    for width_mult in [0.5, 0.75, 1, 1.3]
+        @testset "MobileNetv1" begin
```
darsnack, line 282:
```suggestion
    @testset "MobileNets (width = $width_mult)" for width_mult in [0.5, 0.75, 1, 1.3]
        @testset "MobileNetv1" begin
```

src/layers/mbconv.jl
```diff
@@ -0,0 +1,117 @@
...
+    # squeeze-excite layer
+    if !isnothing(reduction)
+        squeeze_planes = _round_channels((se_from_explanes ? explanes : inplanes) ÷
+                                         reduction, divisor)
+        push!(layers,
+              squeeze_excite(explanes, squeeze_planes; activation, gate_activation = hardσ))
+    end
+    # project
+    append!(layers, conv_norm((1, 1), explanes, outplanes, identity))
+    use_skip = stride == 1 && inplanes == outplanes && !no_skip
```
darsnack, line 97: Is `no_skip` used somewhere?

darsnack (changes requested):
This is looking really clean. I'm quite happy with how this turned out.

src/convnets/builders/mbconv.jl
```diff
@@ -0,0 +1,107 @@
...
+    function get_layers(block_idx::Integer)
+        inplanes = block_idx == 1 ? inplanes : outplanes
+        explanes = _round_channels(inplanes * expansion, 8)
+        stride = block_idx == 1 ? stride : 1
+        block = block_fn((k, k), inplanes, explanes, outplanes, activation; norm_layer,
+                         stride, reduction, no_skip = true, kwargs...)
+        return stride == 1 && inplanes == outplanes ? (identity, block) : (block,)
+    end
+    return get_layers, nrepeats
+end
```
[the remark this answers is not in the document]
theabhirath: Oh yeah, I just was writing a lot of code for some reason here 😂 That is exactly right

src/layers/mbconv.jl
theabhirath: For completeness. We toggle `no_skip` deliberately to `false` inside the builders so that the main CNN builder can add the `Parallel`, but for someone not expecting this behaviour they should be able to use the layer independently and get the `SkipConnections` as intended when the stride is  1 and `inplanes == outplanes`

src/layers/conv.jl
theabhirath: Yeah this makes sense - we might just need to redirect users to a section on using closures from the Julia docs when we write docs for this so that they're comfortable with the syntax needed

src/layers/mbconv.jl
darsnack: But that stride and channel logic isn't part of the paper "theory," right? It's a heuristic for automated building which isn't being done when the layer is used independently. It seems less magical to me for someone who wants to use this independently to explicitly write:
```julia
SkipConnection(mbconv(...), +)
```
And we avoid more keywords switches.

src/layers/conv.jl
darsnack: Yes, but also the revamped docs should have a "how-to" guide on `conv_norm` and friends. Directly spell out this use case then only relying on the linked Julia docs + our docstring.

src/layers/mbconv.jl
theabhirath: > But that stride and channel logic isn't part of the paper "theory," right?

Umm...this is kinda debatable, I guess? MobileNetv2 (which introduces the inverted residual block) talks about the residual connection in the inverted residual block as being key to the design. In fact I was planning on adding these same toggles to the ResNet blocks for completeness since the way I see it a residual block which doesn't return with a residual connection doesn't feel right. We can always toggle this in relevant builders (you're right that it is one more keyword to handle, though)

src/convnets/builders/core.jl
```diff
@@ -0,0 +1,31 @@
+# TODO potentially refactor other CNNs to use this
+function cnn_stages(get_layers, block_repeats::AbstractVector{<:Integer}, connection)
```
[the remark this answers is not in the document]
theabhirath: Hmmm but then we lose the ability to be able to build `MobileNetv1` with this design, for example and it very clearly lends itself to this structure

src/layers/mbconv.jl
darsnack: There are many papers on ResNet style blocks and even more code implementations. So terminology gets all mixed up. But in the original paper, "residual" specifically refers to a single branch of the skip connection. If we want to learn a mapping $H(x)$, then the block learns $H(x) = F(x) + x$ (in the basic identity block) where $F(x) = H(x) - x$ is the "residual," i.e. what _remains_ when $x$ is taken out of $H(x)$.

On a more practical note, I'd prefer to avoid keyword switches for a non-essential task that is trivial to do manually. It's always possible to add these features later if someone wants them, but no point complicating an already complex function prematurely.

src/convnets/builders/core.jl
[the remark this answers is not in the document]
darsnack: We don't lose the ability, we can still do it. We're just not encouraging users to use it for vanilla CNNs.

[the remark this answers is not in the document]
theabhirath: No I mean if we don't define the dispatch used below, we can't create the layers for `MobileNet` unless we pass in something like `connection = nothing`. Because passing in `connection = +` even though it won't be used would be semantically incorrect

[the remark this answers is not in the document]
theabhirath: This is one of the reasons I feel the residuals for all the blocks should be shifted into the block structure and we should used `cnn_stages` only to call `get_layers`. It means that we don't have to worry about the connection at the highest level and let the blocks handle it. It also means that down the line, someone wanting to use a DenseNet-ResNet fusion can do it in one function instead of having to work with two separate `cnn_stages` and then concatenating them. Suddenly all the mid-level model APIs look very uniform, which I think is desirable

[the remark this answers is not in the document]
darsnack: Why can't we pass `connection = nothing`? You could even merge the `@assert` for `isnothing(connection)` and `length(branches) == 1` into the top dispatch. Using multiple dispatch in place of a conditional results in an unnecessary amount of code duplication here.

I could see the `get_layers` returning something like `(connection, (branches...))`, but completely removing the `Parallel` construction into the closure seems too far to me. At that point `cnn_stages` is a vacuous function. It has the appearance of being very powerful because it is used everywhere, but it does not actually do anything. It only automates a doubly nested loop, calling another function that does all the work with the indices of the loop. It is helpful to follow Occam's razor here: we want something as simple as possible but no simpler. You could argue that keeping the `Parallel` within `cnn_stages` does marginally more work. But just that small change makes a big difference...it provides a useful abstraction for thinking about how residual networks are built up. In contrast, an infinitely powerful callback that does it all provides no useful framework.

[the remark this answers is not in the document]
darsnack: In the end, we can keep the name `cnn_stages` but I'd like to limit the duplicated code.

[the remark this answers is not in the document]
theabhirath: > Why can't we pass `connection = nothing`?

I mean we _can_, but then we're implicitly encouraging using this function for non-residual structures in our own source code, which seems like a bit of a contradiction from what we want this function to be. My thought process was that even if we merge the conditional into the top dispatch, by allowing CNNs with one branch to be a valid model structure we are anyway taking on the responsibility of handling certain non-residual structures. And at that point in my opinion having a separate dispatch just makes it easier to reason about residual vs non-residual structures (passing `connection = nothing` is less intuitive than not having a connection at all, in Julian style - Pythonic notation is usually the former because they can't tap into multiple dispatch).

About `cnn_stages` - you're right that having `Parallel` in there provides a semblance of structure. Maybe then the question moves to exploiting the return types from the builder better - how about we return the connection function and the type of connection as well? We'll need an internal helper function that dispatches on say, `Chain`, `Parallel` and `SkipConnection`, maybe, but what this will mean is that we don't lose the structure that I agree is vital to `cnn_stages` being a non-vacuous function while at the same time allowing users the flexibility to construct the model of their liking.

src/layers/conv.jl
theabhirath: Resolved, but leaving this here so that I remember to add docs for this

src/convnets/builders/core.jl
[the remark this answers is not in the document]
darsnack: Okay I wrote a whole response on why I think we shouldn't even allow a no-connection case...but I changed my mind in the end. What I like is your current implementation in terms of API. We still only have the `get_layers` return branches. Returning the type + the connection + the branches seems like just an awkward way to not have the closure call the constructor. It literally is `Parallel, connection, (branches...)`. It is a couple spaces away from just calling the constructor. I think our current design is the right balance between flexibility and clarity.

So the only remaining issue here is how to handle the code duplication. You can use a shadow method:
```julia
function _cnn_stages(get_layers, block_repeats, connection)
# code including the isnothing(connection) assert
end

cnn_stages(get_layers, block_repeats, connection) =
     _cnn_stages(get_layers, block_repeats, connection)
cnn_stages(get_layers, block_repeats) =
     _cnn_stages(get_layers, block_repeats, nothing)
```
Or you can set a default
```julia
function cnn_stages(get_layers, block_repeats, connection = nothing)
# code including the isnothing(connection) assert
end
```
In practice, both will be called the same way. Either pass a connection or don't.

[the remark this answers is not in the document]
theabhirath: Yep, this is done now!

darsnack (approved):
Looks good! Let's wait for CI to turn green.

theabhirath:
Intermittent failures for the ConvNeXt, ConvMixer testset on Julia 1 seem memory related (nightly tests pass. Very funny thing to suddenly start occurring). The pretrained ResNets will be broken until https://github.com/FluxML/Flux.jl/pull/2041 lands. Everything else should pass though

ToucheSir:
CI is now green :) (modulo ResNet note above)

darsnack:
On ImageNet? I used the MobileNets in this package a while back and trained them on CIFAR-100. Looking at that code, I used Adam and logit CE loss. Also a bunch of data augmentation. MobileNets share some code overlap with EfficientNets in the package.

IanButterworth:
I'm trying on a custom dataset but perhaps it makes sense to start with something tried and tested.

darsnack:
Yeah if something like CIFAR-10 doesn't work, then at least someone here can try and reproduce.

IanButterworth:
Switching to `EfficientNet(:b0; ...` I get

Screenshot 2023-12-08 at 3 02 42 PM

darsnack:
Okay I'll make my own script over the weekend and sanity check.

IanButterworth:
Thanks! 

Comparing to this example https://www.kaggle.com/code/paultimothymooney/efficientnetv2-with-tensorflow?scriptVersionId=120655976&cellId=27

- It uses SGD, not Adam.
- Learning rate is 0.005 vs the default Adam 0.001
- Primary momentum is the same
- It uses label smoothing

And performance maxes out within a few epochs

IanButterworth:
@darsnack i was just wondering whether you have had a chance to test this or check my code for any obvious issues. Thanks

IanButterworth:
Using https://github.com/IanButterworth/EfficientNet-Training/commit/ec8008408b49a38042a0c40ed58d27bd460234d2 and switching just the model to `ResNet` gets a much more reasonable result
```
model = ResNet(18; pretrain=false, inchannels=3, nclasses=length(labels)) |> device
```

Screenshot 2023-12-13 at 12 46 27 PM


Other than there being some issue with the model, could it be that EfficientNet requires some subtle difference in the loss function or optimizer setups?

IanButterworth:
`EfficientNet` (not V2)
```
model = EfficientNet(:b0; pretrain=false, inchannels=3, nclasses=length(labels)) |> device
```

Screenshot 2023-12-13 at 12 59 21 PM

theabhirath:
I was trying a lot of these on CIFAR-10 during my GSoC and was facing issues, including with ResNet – the accuracies were nowhere near what I could get with PyTorch, for example. I remember trying to debug this but then had to give up since I got occupied with other commitments. IIRC at the time one theory was that there might be something wrong with the gradients, but we didn't nearly manage to get enough training examples on the GPU to confirm. I could try running these again if I got a GPU somehow 😅

The difference between EfficientNet and ResNet is weird, but not unexpected because they do use different kinds of layers. Maybe MobileNet vs EfficientNet tells us more? The underlying code over there is the exact same because of the way the models are constructed. Even for ResNet, though, the loss curve looks kinda bumpy...

IanButterworth:
Interesting. 

I was considering setting up a job to do the same for all appropriate models in Metalhead on my machine, and tracking any useful metrics while doing so for a comparison table. I have a relatively decent setup so it might not take too long. 

I can understand that full training isn't a part of CI because of resources, but I think it needs to be demonstrated that these models are trainable somewhere public.

IanButterworth:
I've started putting together a benchmarking script here. https://github.com/FluxML/Metalhead.jl/pull/264
It'd be great to get feedback on it.

ToucheSir:
As @theabhirath mentioned, this has come up before but we never got to the bottom of why because of time constraints. If you have some bandwidth to do some diagnostic runs looking at whether gradients, activations etc are misbehaving and where they're misbehaving, we could narrow the problem down to something actionable. Whether that be a bug in Metalhead itself or something further upstream like Flux, Zygote or CUDA.

IanButterworth:
Ok. I think my minimum goal then is to characterise the issue (is it all models) and make it clear in the docs which ones cannot be expected to be trainable. 

As someone coming to this package to use it to train on a custom dataset it's been quite a time sync just to get to that understanding.

ToucheSir:
If it's a either-or deal between that and finding out why a specific model (e.g. EfficientNet) has problems, I'd lean towards the latter. But if the goal is to do both, that sounds good to me.

IanButterworth:
I lean more towards what I'm currently trying because I'm not sure I have the knowledge/skill set to dive in and debug. 

Tbh if I find a model that trains and is performant I may declare victory, but share findings. 

Or maybe I strike lucky in a dive.

darsnack:
Sorry for the late reply. I did start writing a script, but I never gotten around to starting my testing. I kept meaning to reply as soon as I did that, but it's been a busy week.

Looks like we might have some duplicate work. Either way, I uploaded my code to a repo [here](https://github.com/darsnack/metalhead-bench). I've add all of you as collaborators. The biggest help right now would be is someone can add in all the remaining models + configs to the `main.jl` script. Output from the script is being logged [here](https://wandb.ai/darsnack/metalhead-bench/). It should be a public wandb workspace that we can all view.

darsnack:
Based on the current runs that finished, it looks like all the EfficientNetv2 models have some bug. Only the `:small` variant trains to completion. The rest all drop to NaN loss at some point, with the larger variants dropping to NaN in the first epoch.

ResNets have the quirk Abhirath noticed during the summer where the models starts off fine then start to overfit to the training data.

I also have AlexNet and EfficientNet (v1) queued up. Let's see how those do.

darsnack:
I will say though that the ResNet loss curves don't look as bad as I remember them. Perhaps in this case, a different learning rate would fix things.

ToucheSir:
My recollection was that the PyTorch resnets converged faster and overfit less easily even with less help from data augmentation. Is it straightforward to do a head-to-head comparison?

ToucheSir:
I'm able to replicate the poor training behaviour of EfficientNet-B0 on a local machine, which happens to have an AMD GPU. this suggests the problem may not lie with anything CUDA-specific.

IanButterworth:
Great, thanks. How does one go about debugging this kind of thing? Are there generic tools for visualizing the model that could help?

darsnack:
I'll modify the script to log gradient norms by layer and also do some local debugging just to sanity check the output isn't obviously wrong.

I'll also add MobileNet to the script. I think that might be a good reference model to compare against assuming it does converge. If it works, that would narrow the issue down to the specific "inverted residual block" that EfficientNet uses.

ToucheSir:
If either of you have a machine with a decently powerful CPU, I think a CPU-only run would be interesting to see if we can isolate GPU libraries as a possibility altogether.

IanButterworth:
I'm giving EfficientNetv2 a go with cpu and `-t16`. Looks like each epoch will take 3 hours
Screenshot 2023-12-16 at 8 58 09 PM
```
julia> versioninfo()
Julia Version 1.9.4
Commit 8e5136fa297 (2023-11-14 08:46 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen 9 5950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver3)
  Threads: 16 on 32 virtual cores
```

ToucheSir:
I was seeing less time than that for b0 on my decidedly less powerful machine, so maybe using a smaller size would help? We've established that all EfficientNets are affected after all.

theabhirath:
Could we shift the discussion to a GitHub discussion? I'd hate to see this getting lost in a closed PR 😅

IanButterworth:
EfficientNet B0 after 15 epochs
Screenshot 2023-12-17 at 8 51 11 AM

IanButterworth:
I've updated progress on my benchmark script here https://github.com/FluxML/Metalhead.jl/pull/264#issuecomment-1866823433

Would anyone be able to help me get the errors resolved? I guess most of them are input image size issues?

theabhirath:
> I've updated progress on my benchmark script here [#264 (comment)](https://github.com/FluxML/Metalhead.jl/pull/264#issuecomment-1866823433)
> 
> Would anyone be able to help me get the errors resolved? I guess most of them are input image size issues?

I've commented there! Also just to understand your previous EfficientNet training on CPU graph, it still seems to have those fluctuations in loss, right? Meaning that the issue might not be exactly GPU linked (or at least not _only_ GPU linked).

darsnack:
Updated results: https://wandb.ai/darsnack/metalhead-bench/

For me, I see MobileNetv1 and MobileNetv2 succeeding while MobileNetv3 fails. I tried to check what the differences are by inspection, and the one major different was the use of `hardswish` and `hardsigmoid`.

So, I ran another MobileNetv3 but with `hardswish` and `hardsigmoid` replaced by `relu`. Now, it trains to a reasonable accuracy relative to the others. I will re-run again on the CPU to double check.

PR #200, opened 2022-08-26
Unify  lower level API for EfficientNet and MobileNet model families

theabhirath:
This PR unifies the lower level API for the EfficientNet and MobileNet model families into a single `irmodelbuilder` function. This followed as a natural consequence of #198 - a considerable amount of the calculations, model structure and configurations have now been abstracted out enough into pieces that compose, meaning that most of the models in these two families can be almost completely identified by their configuration dict and a couple of other arguments. Amongst other things, what this meant was 
1. a simplification of the builders for the inverted residual block stages,
2. an ability to expose a more friendly mid-level API for all models that doesn't need the user to specify unexported, constant configuration dicts. This step is still in process but once completed, should hopefully result in a smoother experience for advanced users and
3. Even more reduction of code duplication

## Documentation TODO

- [x] Builders
- [x] Mid level API

## Miscellaneous Notes

In a step closer to a uniform API, the higher level model APIs now all accept the `pretrain` keyword argument.


src/layers/selayers.jl
```diff
@@ -21,19 +21,13 @@ Creates a squeeze-and-excitation layer used in MobileNets, EfficientNets and SE-
...
-        norm_layer(inplanes),
-        gate_activation]
-    return SkipConnection(Chain(filter!(!=(identity), layers)...), .*)
+function squeeze_excite(inplanes::Integer, squeeze_planes::Integer; norm_layer = identity,
+                        activation = relu, gate_activation = sigmoid)
+    return SkipConnection(Chain(AdaptiveMeanPool((1, 1)),
+                                conv_norm((1, 1), inplanes, squeeze_planes, activation;
+                                          norm_layer)...,
+                                conv_norm((1, 1), squeeze_planes, inplanes,
+                                          gate_activation; norm_layer)...), .*)
```
ToucheSir, line 30: TIL that `(.*)` returns a `Base.Broadcast.BroadcastFunction`. Maybe we should create an alias for it in NNlib...

src/convnets/builders/resnet.jl
```diff
@@ -0,0 +1,9 @@
+function resnetbuilder(img_dims, stem, get_layers, block_repeats::AbstractVector{<:Integer},
```
ToucheSir, line 1: It's a small thing, but I wonder if these functions which actually return a `Chain` instead of a builder function should be named with a leading imperative. e.g.
```suggestion
function build_resnet(img_dims, stem, get_layers, block_repeats::AbstractVector{<:Integer},
```
Then there is a separation between the functions which give you back something to call later, and the ones which give you a model immediately.

src/convnets/builders/mbconv.jl
```diff
@@ -12,13 +17,14 @@ function dwsepconv_builder(block_configs, inplanes::Integer, stage_idx::Integer,
                                stride, pad = SamePad(), norm_layer, kwargs...)...)
         return (block,)
     end
-    return get_layers, nrepeats
+    return get_layers, ceil(Int, nrepeats * depth_mult)
 end
+_get_builder(::typeof(dwsep_conv_norm)) = dwsepconv_builder
```
ToucheSir, line 22: Perhaps this is what your comment above is alluding to, but as-is it seems like `_get_builder(::typeof(f))` and `f_builder` could be merged into one function.

ToucheSir (commented):
Fewer comments than I thought for a first pass :)

src/convnets/builders/mbconv.jl
theabhirath: They _could_, but then the config dict has to specify the builder function and not the block function

src/convnets/builders/resnet.jl
theabhirath: Makes sense. Names _are_ so important here because they really should try and convey the point of the function even before the user starts reading the docs

src/convnets/builders/mbconv.jl
ToucheSir: Not sure I follow? What I meant was that the function currently called `dwsepconv_builder(args...)` could just be `_get_builder(::typeof(dwsep_conv_norm), args...)`. AFAICT this should not change what is stored in the config dict because it simply eliminates one function dispatch from the call chain.

theabhirath: Ah, I didn't think about eliminating the explicit builder altogether and just making it dispatch. That makes an awful lot of sense. I still keep forgetting multiple dispatch exists sometimes

ToucheSir: Well, you _are_ using it already for `_get_builder`! This merging of functions just happens to be something that came out of the wash while refactoring the `mbconv_stage_builder`.

theabhirath:
This PR now also adds an additional `batch` mode to `StochasticDepth` (renamed from `DropPath` because it's not very descriptive and also to mimic torchvision) for completeness, fixes some minor issues with linear scheduling rates, renames all the `*_rate` to `*_prob` to make it clear that these are probability values and finally, adds `StochasticDepth` with a default value of 0.2 to the EfficientNets, finally achieving feature parity with the papers and torchvision/timm

src/convnets/builders/mbconv.jl
theabhirath: Yep, this is done now - perhaps the only nitpick is perhaps that to allow this kind of structure there is a redundant `stochastic_depth_prob` argument now being passed in to the depthwise separable convolution builder. But it's a small price to pay for uniform code

src/layers/drop.jl
```diff
@@ -126,36 +121,39 @@ function Base.show(io::IO, d::DropBlock)
     return print(io, ")")
 end
 
-# TODO look into "row" mode for stochastic depth
 """
-    DropPath(p; [rng = rng_from_array(x)])
```
ToucheSir, line 131: To make this non/less breaking, adding a deprecation for `DropPath` would work.

theabhirath: But the Layers API is unexported, so does this still count as a breaking change?

theabhirath: It really should be exported in 0.8, but for now it isn't so I suppose it might work out in our favour 😄

ToucheSir: I assumed it was why you added the label, but that is true. Built docs list it as "public", but I think that's a false positive and JuliaHub tells me no registered packages are using it.

theabhirath: Oh no, the breaking change is that `EfficientNet` literally returns a different model 😅

theabhirath:
This PR is complete I think - any further docs improvements would be best handled in a separate PR after landing this one and then #199

darsnack (changes requested):
Just a few final naming things. Also, we can tackle more thorough docs including docstrings separately.

src/convnets/builders/mbconv.jl
```diff
@@ -1,9 +1,23 @@
...
+
+Constructs a collection of inverted residual blocks for a given stage. Note that
+this function is not intended to be called directly, but rather by the [`mbconv_stage_builder`](@ref)
+function. This function must only be extended if the user wishes to extend a custom inverted
+residual block type.
+"""
+function irblockbuilder(::typeof(dwsep_conv_norm), block_configs::AbstractVector{<:Tuple},
+                        inplanes::Integer, stage_idx::Integer, scalings::NTuple{2, Real};
+                        stochastic_depth_prob = nothing, norm_layer = BatchNorm,
+                        divisor::Integer = 8, kwargs...)
```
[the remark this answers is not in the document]
theabhirath: These are being used though, to pass stuff like the `se_round_fn` in for `MobileNetv3`

theabhirath:
Bump?

src/convnets/builders/mbconv.jl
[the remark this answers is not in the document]
darsnack: I specifically mean stuff like `stochastic_depth_rate`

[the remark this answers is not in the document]
theabhirath: Oh you mean I can simply keep them in `kwargs`?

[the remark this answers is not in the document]
theabhirath: > I specifically mean stuff like `stochastic_depth_prob`

This particular one needs to be there because it will cause a `MethodError` for `dwsep_conv_norm` if passed through

darsnack (approved):

src/convnets/builders/mbconv.jl
```diff
@@ -29,79 +46,52 @@ function mbconv_builder(block_configs, inplanes::Integer, stage_idx::Integer,
...
 
-function fused_mbconv_builder(block_configs, inplanes::Integer, stage_idx::Integer;
-                              norm_layer = BatchNorm, kwargs...)
-    block_fn, k, outplanes, expansion, stride, nrepeats, activation = block_configs[stage_idx]
+function invresbuilder(::typeof(fused_mbconv), block_configs::AbstractVector{<:Tuple},
+                       inplanes::Integer, stage_idx::Integer, scalings::NTuple{2, Real};
+                       stochastic_depth_prob = nothing, norm_layer = BatchNorm,
+                       divisor::Integer = 8, kwargs...)
+    width_mult, depth_mult = scalings
+    block_repeats = [ceil(Int, block_configs[idx][end - 1] * depth_mult)
```
shivance, line 73: For Mobilenet the config list is 
```
const MOBILENETV2_CONFIGS = [
    (mbconv, 3, 16, 1, 1, 1, nothing, relu6),
    (mbconv, 3, 24, 6, 2, 2, nothing, relu6),
    (mbconv, 3, 32, 6, 2, 3, nothing, relu6),
    (mbconv, 3, 64, 6, 2, 4, nothing, relu6),
    (mbconv, 3, 96, 6, 1, 3, nothing, relu6),
    (mbconv, 3, 160, 6, 2, 3, nothing, relu6),
    (mbconv, 3, 320, 6, 1, 1, nothing, relu6),
]
```

so won't `block_repeats` just be a list of `nothing` ?
and as the name suggest, it tells how many time a particular block of layers is repeated?

PR #202, opened 2022-09-07
Update buildkite pipeline

darsnack:
This updates the GPU CI to test on LTS + latest.


.buildkite/pipeline.yml
```diff
@@ -1,10 +1,37 @@
...
+      - JuliaCI/julia#v1:
+          # Drop default "registries" directory, so it is not persisted from execution to execution
+          # Taken from https://github.com/JuliaLang/julia/blob/v1.7.2/.buildkite/pipelines/main/platforms/package_linux.yml#L11-L12
+          persist_depot_dirs: packages,artifacts,compiled
+          version: "1.6"
+      - JuliaCI/julia-test#v1: ~
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
+    timeout_in_minutes: 60
```
ToucheSir, line 13: This is probably not enough time, but we'll see. Would it make sense to only run some suites?

darsnack: Probably running just the layers is sufficient, but I haven't tried ReTest on Buildkite.

theabhirath: > We also don't separately test the layers.

I'm working on this - this might land after the GSoC, but I have a couple of specific tests in mind for the custom layers that I've bookmarked to go back and implement in Metalhead

ToucheSir: Thoughts on just running the layers tests for now so we can get this in?

CarloLucibello:
superseded by #222

Issue #196, opened 2022-08-04
Tweak GoogLeNet and Inception family to match the torchvision implementations 


theabhirath:
These changes will make it easier to port over pretrained weights for the models from PyTorch.

- [x] Right now, GoogLeNet matches the implementation in the paper, which does not use batch normalisation. However, torchvision uses [batch normalisation layers after the convolution layers](https://github.com/pytorch/vision/blob/6db1569c89094cf23f3bc41f79275c45e9fcb3f3/torchvision/models/googlenet.py#L285-L295) in both the inception block and the stem. It also uses `bias = false` for the convolution layer and `eps = 0.001` for the batch normalisation layer. We could decide if similar to VGG, we can have a toggle for the batch normalisation in the model. (closed by #205)
- [x] InceptionV3, InceptionV4 and InceptionResNetv2 have the right placement of normalisation layers but similar to GoogLeNet, use [`bias = false` for the convolution layer and `eps = 0.001` for the batch normalisation layer.](https://github.com/pytorch/vision/blob/96aa3d928c6faaf39defd846966421679244412d/torchvision/models/inception.py#L398-L407) (should be closed by #198)

darsnack:
A toggle would be a good idea so that the paper variant is still easily available.

pri1311:
Hi, this issue seems like a good place to start contributing.
`
function convolution(kernel_size, inplanes, outplanes, batchNorm; kwargs...)
    if batchNorm
        return basic_conv_bn(kernel_size, inplanes, outplanes; kwargs...)
    else 
        return Conv(kernel_size, inplanes => outplanes; kwargs...)
    end
end
`
would implementing such a function and the calling it like this
`convolution((7, 7), inchannels, 64, batchNorm; stride = 2, pad = 3, bias = bias`
be a good way to refactor code for the particular issue?
Or do you have something else in mind?

darsnack:
We already have support for turning off the batch norm [here](https://github.com/FluxML/Metalhead.jl/blob/88bcacb81b643066adecdc9bb108a6dbcaad9e3c/src/layers/conv.jl#L15-L17). The remaining task on this issue it to update the code for GoogLeNet to use that functionality.

pri1311:
Aahh got it! thanks.

PR #205, opened 2022-11-16
Tweak GoogLeNet to match the torchvision implementations

pri1311:
Changed the `Conv` layers to `conv_norm` layers. 

However, I couldn't find a way to specify eps for `BatchNorm`. The default value that flux uses is 1f-5, and torchvision version of GoogleNet has eps set to 0.001.

### PR Checklist

- [x] Tests are added
- [x] Documentation, if applicable

Also ps: Unrelated, but the link to 'contributing docs' in the readme seems to be broken.

### DIFF
--- a/src/convnets/inceptions/googlenet.jl
+++ b/src/convnets/inceptions/googlenet.jl
@@ -16,14 +16,21 @@ Create an inception module for use in GoogLeNet
   - `batchnorm`: set to `true` to include batch normalization after each convolution
   - `bias`: set to `true` to use bias in the convolution layers.
 """
-function inceptionblock(inplanes, out_1x1, red_3x3, out_3x3, red_5x5, out_5x5, pool_proj, batchnorm, bias)
-    branch1 = Chain(basic_conv_bn((1, 1), inplanes, out_1x1; batchnorm = batchnorm, bias = bias)...)
-    branch2 = Chain(basic_conv_bn((1, 1), inplanes, red_3x3; batchnorm = batchnorm, bias = bias)...,
-                    basic_conv_bn((3, 3), red_3x3, out_3x3; batchnorm = batchnorm, bias = bias, pad = 1)...)
-    branch3 = Chain(basic_conv_bn((1, 1), inplanes, red_5x5; batchnorm = batchnorm, bias = bias)...,
-                    basic_conv_bn((5, 5), red_5x5, out_5x5; batchnorm = batchnorm, bias = bias, pad = 2)...)
+function inceptionblock(inplanes, out_1x1, red_3x3, out_3x3, red_5x5, out_5x5, pool_proj,
+                        batchnorm, bias)
+    branch1 = Chain(basic_conv_bn((1, 1), inplanes, out_1x1; batchnorm = batchnorm,
+                                  bias = bias)...)
+    branch2 = Chain(basic_conv_bn((1, 1), inplanes, red_3x3; batchnorm = batchnorm,
+                                  bias = bias)...,
+                    basic_conv_bn((3, 3), red_3x3, out_3x3; batchnorm = batchnorm,
+                                  bias = bias, pad = 1)...)
+    branch3 = Chain(basic_conv_bn((1, 1), inplanes, red_5x5; batchnorm = batchnorm,
+                                  bias = bias)...,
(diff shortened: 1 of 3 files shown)


ToucheSir:
The `BatchNorm` constructor (dee https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.BatchNorm) does have an option to specify an epsilon (it's spelled with the symbol `ϵ` instead of the abbreviation `eps`), but it does not seem like `conv_norm` has an option to pass any additional args through to the norm layer it creates. I think it would be fine to add an extra keyword arg to `conv_norm` for this, but maybe others have different proposals.

theabhirath:
> However, I couldn't find a way to specify eps for `BatchNorm`. The default value that flux uses is 1f-5, and torchvision version of GoogleNet has eps set to 0.001.

We have [`basic_conv_bn`](https://github.com/FluxML/Metalhead.jl/blob/88bcacb81b643066adecdc9bb108a6dbcaad9e3c/src/layers/conv.jl#L64-L85) which already does this. This is what is currently being used for the other Inception models.

> I think it would be fine to add an extra keyword arg to `conv_norm` for this, but maybe other maintainers have different proposals.

I'd tried this during the refactor, but it added too much clutter to the docstring and I had to handle other norm layers manually (although for `eps` in particular that might not be a problem).

darsnack:
> it does not seem like conv_norm has an option to pass any additional args through to the norm layer it creates

I think the rational here is that conv will have keywords customized more than batch norm, so it gets preference for pass-through keywords. This avoids having different subsets of keywords going to different layers which I think is confusing. The normalization layer then has to use the slightly more verbose syntax of a closure: `(planes, act) -> BatchNorm(planes, act; ϵ = 1f-3)`.

Though as Abhirath mentioned, using `basic_conv_bn` is preferred here.

pri1311:
Would doing this work? - [`norm_layer = batchnorm ? (args...; kwargs...) -> BatchNorm(args...; ϵ = 1.0f-3) : identity`](https://github.com/FluxML/Metalhead.jl/pull/205/files#diff-d4ad39cff485d98330ceb76b0fd8975e4b0c8c7a5119c02fb9bb4c04a6ed0f5cR47)

Because `basic_conv_bn` will require if else statements for every `conv` block.
Or a wrapper function like this - https://github.com/FluxML/Metalhead.jl/issues/196#issuecomment-1316035168

theabhirath:
> Because `basic_conv_bn` will require if else statements for every `conv` block.

I didn't quite follow. `basic_conv_bn` calls `conv_norm` under the hood, so this should just be a matter of adding a toggle to `basic_conv_bn` to switch off the batch normalisation (and pass in `identity` to the `norm_layer` argument in `conv_bn` for that case). So the wrapper function you are suggesting _is_ in fact, `basic_conv_bn`. What you've done in the PR works too, but the reason I introduced `basic_conv_bn` was so that the Inception family could have one function that handled the setting of these values so that it's clear to the users that they're related.

pri1311:
So you are suggesting tweaking the `basic_conv_bn` function to have the toggle, right? 

```
function basic_conv_bn(kernel_size::Dims{2}, inplanes, outplanes, activation = relu; batchnorm::Bool = true
                       kwargs...)
    # TensorFlow uses a default epsilon of 1e-3 for BatchNorm
    norm_layer = batchnorm ? (args...; kwargs...) -> BatchNorm(args...; ϵ = 1.0f-3, kwargs...) : identity
    return conv_norm(kernel_size, inplanes, outplanes, activation; norm_layer, kwargs...)
end
```

theabhirath:
Yep, that seems perfect!

pri1311:
Thank you for all the help @theabhirath! 

I had one more question though, the current implementation of GoogLeNet does not have `relu` activations after Conv blocks, but to my knowledge, the paper mentions having relu activations after convolutions as well torchvision has relu activation. Should I make that change too?

theabhirath:
> I had one more question though, the current implementation of GoogLeNet does not have `relu` activations after Conv blocks, but to my knowledge, the paper mentions having relu activations after convolutions as well torchvision has relu activation. Should I make that change too?

Sure, go ahead! The closer we are to paper parity the better 😄

theabhirath (approved):
LGTM! Could you just run the formatter locally once and commit so that the alignment/spacing for the code matches the rest of the repository? Otherwise everything seems great, thank you so much for the contribution!

Issue #112, opened 2022-02-07
Add model implementations


darsnack:
Below are models that we still need to implement

- [ ] Inception v4 (https://arxiv.org/abs/1602.07261v2)
- [ ] Inception ResNet v2 (https://arxiv.org/abs/1602.07261v2)
- [ ] Xception (https://openaccess.thecvf.com/content_cvpr_2017/html/Chollet_Xception_Deep_Learning_CVPR_2017_paper.html)
- [x] ResNeXt
- [x] MobileNet v1
- [x] MobileNet v2
- [x] MLPMixer
- [ ] ViT

darsnack:
cc @theabhirath please add the models you are planning on implementing to this list

darsnack:
A starting point for EfficientNet can be found at https://github.com/pxl-th/EfficientNet.jl (see #106).

theabhirath:
I'm planning on  

- [ ] [CaiT](https://arxiv.org/abs/2103.17239)
- [ ] [LeViT(https://arxiv.org/abs/2104.01136)
- [ ] [Swin](https://arxiv.org/abs/2103.14030)
- [ ] [MobileViT](https://arxiv.org/abs/2110.02178)
- [ ] [ConvNeXt](https://arxiv.org/abs/2201.03545)

at first. I also want to start on object detection and semantic segmentation models but there's a lot of helper functions related to the fields that I will have to write - if everything CV-related is planned to be in Metalhead, then I'll go ahead and start coding them up

rutvij-25:
Hello, Can I work on inception v4 and efficient net?(Also I have already coded ESRGANs)

DhairyaLGandhi:
There's #113 for efficient net. Would be good to port srgan etc into Metalhead as well

Alexander-Barth:
I could provide the pretrained weights for VGG and ResNet converted from PyTorch (once some minor changes to the Metalhead models are merged so that they are equivalent to the PyTorch model).

(I would be very interested in an ESRGAN implementation. 😃)

DhairyaLGandhi:
Also see #109 for pretrained models. It's hard to know before hand which weights would work well between pytorch and Flux, but if we have some pretrained weights we can validate, that would be welcome!

darsnack:
> Hello, Can I work on inception v4 and efficient net?(Also I have already coded ESRGANs)

As mentioned, there is already a PR for EfficientNet, but InceptionNet v4 would be very welcome!

[...]

PR #137, opened 2022-03-21
Added the FastAI UNet implementation

ceferisbarov:
Following [this](https://github.com/FluxML/Metalhead.jl/pull/136) discussion, I closed that PR and opened this one where I have added the [FastAI implementation](https://github.com/FluxML/FastAI.jl/blob/67ab2ffd5170ff7e3b2efeb3609f5ab981990741/src/Vision/models/unet.jl) of UNet.   
  
The model itself seems to have some problems (they exist in FastAI package, too). I will try to clarify and solve them.

PR #210, opened 2022-12-27
Adding UNet implementation

shivance:
This PR adds the [UNet](https://arxiv.org/abs/1505.04597v1) implementation to Metalhead.jl in favor of #112 

I've referred official torchhub implementation [here](https://github.com/mateuszbuda/brain-segmentation-pytorch) and @DhairyaLGandhi 's [UNet.jl](https://github.com/DhairyaLGandhi/UNet.jl) package.

 

### PR Checklist

- [x] Tests are added
- [x] Documentation, if applicable


shivance:
The PR is ready for code review

ToucheSir:
Metalhead has a JuliaFormatter config, so if your editor supports that I would recommend running it to help with code style adherence.

pri1311 (commented):
I doubt there is supposed to be Batchnorm and ReLu layer before the concat + conv layer

shivance:
@ToucheSir I ran [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) using `format(".")` 
Looks like the other files in Metalhead were not formatted as well.
Should I keep it this way or just format unet.jl ?

@pri1311 yup ! corrected that.

src/Metalhead.jl
```diff
@@ -2,6 +2,7 @@ module Metalhead
 
 using Flux
 using Flux: Zygote, outputsize
+using Distributions: Normal
```
pri1311, line 5: Did JuliaFormatter add the new dependency?

src/convnets/unet.jl
```diff
@@ -0,0 +1,83 @@
...
+    out = u.bottleneck(out)
+
+    for i in 4:-1:1
+        out = u.upconv[5 - i](out)
+        out = cat(out, enc_out[i]; dims = 3)
+        out = u.decoder[i](out)
+    end
+
+    return σ(u.final_conv(out))
+end
```
darsnack, line 83: Instead of custom forward pass, can we construct this using `Parallel` from Flux?

src/utilities.jl
```diff
@@ -79,3 +79,9 @@ linear_scheduler(drop_prob::Nothing; depth::Integer) = fill(drop_prob, depth)
 function _checkconfig(config, configs)
     @assert config in configs "Invalid configuration. Must be one of $(sort(collect(configs)))."
 end
+
+# Utility function to be used for generating random normal distribution used for 
+# initializing layer parameters
+function _random_normal(shape...)
+    return Float64.(rand(Normal(0.0f0, 0.02f0), shape...))
+end
```
pri1311, line 87: Why the random_normal initialization specifically? I don't think the paper mentions anything in particular. This adds a dependency on Distributions.jl which didn't exist before I believe.

ToucheSir: If this doesn't exist in Flux already (I would be surprised if it didn't), then it should be added there and not in Metalhead.

src/convnets/unet.jl
ToucheSir: Either way, the custom forward pass should not use loops as Zygote doesn't do well with them.

ToucheSir: Either way, the custom forward pass should not use loops as Zygote doesn't do well with them.

src/utilities.jl
pri1311: Maybe using [`Flux.randn32`](https://fluxml.ai/Flux.jl/dev/utilities/#Flux.randn32)?

shivance:
@ToucheSir @darsnack Finally made it.
Successfully modified the model to use only Parallel and not use custom forward pass.

It was both challenging & confusing simultaneously as I was continuously getting dimension mismatch error. I knew it very well that it's because of Parallel, and the tensors were propagating through the maxpool which it wasn't supposed to.

I tried debugger as well, but there seemed some problem with it.

Following helped :
1. I drew architecture on paper, and wrote down all layers, and matched with my code.
2. Then I localized the error.

I was successfully able to resolve the error by moving

`layers = Chain(layers, decoder_layer) `

to before the decoder block. I realized that I have been chaining the layers with decoder, (whilst decoder layers are Chain of concat and decoder conv layers). This was causing error as chaining at this stage would make all tensors flow through concat again, thus dimensionmismatch.

Moving it before avoided this case.

src/convnets/unet.jl
```diff
@@ -0,0 +1,82 @@
...
+                 norm2 = BatchNorm(out_chs, relu))
+end
+
+function upconv_block(in_chs::Int, out_chs::Int, kernel = (2, 2))
+    return ConvTranspose(kernel, in_chs => out_chs; stride = (2, 2))
+end
+
+function cat_fn(layers...)
+    return cat(layers...; dims = 3)
+end
```
darsnack, line 15: There should be `cat_channels` already in Metalhead.

src/convnets/unet.jl
```diff
@@ -0,0 +1,82 @@
...
+
+function unet(in_channels::Integer = 3, out_channels::Integer = in_channels,
+              features::Integer = 32)
+    encoder_conv_layers = []
+    append!(encoder_conv_layers,
+            [unet_block(in_channels, features)])
```
darsnack, line 21:
```suggestion
    push!(encoder_conv_layers,
          unet_block(in_channels, features))
```

src/convnets/unet.jl
```diff
@@ -0,0 +1,82 @@
...
+            [unet_block(features * 2^i, features * 2^(i + 1)) for i in 0:2])
+
+    encoder_conv = Chain(encoder_conv_layers)
+    encoder_pool = [Chain(encoder_conv[i], MaxPool((2, 2); stride = (2, 2))) for i in 1:4]
+
+    bottleneck = unet_block(features * 8, features * 16)
+    layers = Chain(encoder_conv, bottleneck)
+
+    upconv = Chain([upconv_block(features * 2^(i + 1), features * 2^i)
+                    for i in 0:3])
```
darsnack, line 33: These lines and similar ones below should have splatting to create a `Chain{Tuple}` instead of a `Chain{Vector}`.

```suggestion
    upconv = Chain([upconv_block(features * 2^(i + 1), features * 2^i)
                    for i in 0:3]...)
```

test/convnets.jl
```diff
@@ -6,120 +6,114 @@
...
+    @test size(m(x_224)) == (1000, 1)
+    if (VGG, sz, bn) in PRETRAINED_MODELS
+        @test acctest(VGG(sz; batchnorm = bn, pretrain = true))
+    else
+        @test_throws ArgumentError VGG(sz, batchnorm = bn, pretrain = true)
     end
-end
+    @test gradtest(m, x_224)
+    _gc()
+end end
```
darsnack, line 21: Has anyone figured out how to stop the formatter from doing this?

darsnack (changes requested):
Great job! I still need to review the architecture details, but here are some initial minor changes.

src/Metalhead.jl
pri1311, line 5:
```suggestion
```

test/convnets.jl
pri1311: Nope, the issue is still open - https://github.com/domluna/JuliaFormatter.jl/issues/604
and as of now, I don't see any [option](https://github.com/domluna/JuliaFormatter.jl/blob/master/src/options.jl) in JuliaFormatter to turn it off.

shivance:
Gtg for next round of review @darsnack @ToucheSir @pri1311 !

src/convnets/alexnet.jl
```diff
@@ -42,7 +42,7 @@ Create a `AlexNet`.
   - `nclasses`: the number of output classes
 
 !!! warning
-
```
ToucheSir, line 45: There are a number of files with unrelated changes like this that make the review difficult. Although removing trailing whitespace is always good, it's better left as its own PR. For this one, can you revert the changes to unrelated files so that we can focus on directly relevant ones?

pri1311:
Following is the output I get from loading the model:
```
UNet(
  Chain(
    Chain(
      Chain([
        Chain(
          conv1 = Conv((3, 3), 3 => 32, pad=1),  # 896 parameters
          norm1 = BatchNorm(32, relu),  # 64 parameters, plus 64
          conv2 = Conv((3, 3), 32 => 32, pad=1),  # 9_248 parameters
          norm2 = BatchNorm(32, relu),  # 64 parameters, plus 64
        ),
        Chain(
          conv1 = Conv((3, 3), 32 => 64, pad=1),  # 18_496 parameters
          norm1 = BatchNorm(64, relu),  # 128 parameters, plus 128
          conv2 = Conv((3, 3), 64 => 64, pad=1),  # 36_928 parameters
          norm2 = BatchNorm(64, relu),  # 128 parameters, plus 128
        ),
        Chain(
          conv1 = Conv((3, 3), 64 => 128, pad=1),  # 73_856 parameters
          norm1 = BatchNorm(128, relu),  # 256 parameters, plus 256
          conv2 = Conv((3, 3), 128 => 128, pad=1),  # 147_584 parameters
          norm2 = BatchNorm(128, relu),  # 256 parameters, plus 256
        ),
        Chain(
          conv1 = Conv((3, 3), 128 => 256, pad=1),  # 295_168 parameters
          norm1 = BatchNorm(256, relu),  # 512 parameters, plus 512
          conv2 = Conv((3, 3), 256 => 256, pad=1),  # 590_080 parameters
          norm2 = BatchNorm(256, relu),  # 512 parameters, plus 512
        ),
      ]),
      Chain(
        conv1 = Conv((3, 3), 256 => 512, pad=1),  # 1_180_160 parameters
        norm1 = BatchNorm(512, relu),   # 1_024 parameters, plus 1_024
        conv2 = Conv((3, 3), 512 => 512, pad=1),  # 2_359_808 parameters
        norm2 = BatchNorm(512, relu),   # 1_024 parameters, plus 1_024
      ),
    ),
    Chain(
      Chain(
        conv1 = Conv((3, 3), 512 => 256, pad=1),  # 1_179_904 parameters
        norm1 = BatchNorm(256, relu),   # 512 parameters, plus 512
        conv2 = Conv((3, 3), 256 => 256, pad=1),  # 590_080 parameters
        norm2 = BatchNorm(256, relu),   # 512 parameters, plus 512
      ),
      Chain(
        conv1 = Conv((3, 3), 256 => 128, pad=1),  # 295_040 parameters
        norm1 = BatchNorm(128, relu),   # 256 parameters, plus 256
        conv2 = Conv((3, 3), 128 => 128, pad=1),  # 147_584 parameters
        norm2 = BatchNorm(128, relu),   # 256 parameters, plus 256
      ),
      Chain(
        conv1 = Conv((3, 3), 128 => 64, pad=1),  # 73_792 parameters
        norm1 = BatchNorm(64, relu),    # 128 parameters, plus 128
        conv2 = Conv((3, 3), 64 => 64, pad=1),  # 36_928 parameters
        norm2 = BatchNorm(64, relu),    # 128 parameters, plus 128
      ),
      Chain(
        conv1 = Conv((3, 3), 64 => 32, pad=1),  # 18_464 parameters
        norm1 = BatchNorm(32, relu),    # 64 parameters, plus 64
        conv2 = Conv((3, 3), 32 => 32, pad=1),  # 9_248 parameters
        norm2 = BatchNorm(32, relu),    # 64 parameters, plus 64
      ),
    ),
  ),
)         # Total: 72 trainable arrays, 7_069_152 parameters,
          # plus 36 non-trainable, 5_888 parameters, summarysize 27.002 MiB.
```

I believe layers are missing

src/convnets/unet.jl
```diff
@@ -0,0 +1,73 @@
+function unet_block(in_chs::Int, out_chs::Int, kernel = (3, 3))
+	return Chain(conv1 = Conv(kernel, in_chs => out_chs; pad = (1, 1)),
+		norm1 = BatchNorm(out_chs, relu),
+		conv2 = Conv(kernel, out_chs => out_chs; pad = (1, 1)),
+		norm2 = BatchNorm(out_chs, relu))
```
pri1311, line 5: Any specific reason for using named layers? I don't see it being used anywhere. I haven't seen Metalhead use such a code convention, so a seems a little inconsistent with the code base.

shivance:
I just checked through outputs which looked fine to me, but now that you point out @pri1311 , I realize that upper layers aren't being chained at all.

I doubt that using parallel isn't possible here. We need custom forward pass.
@darsnack Please review.

src/convnets/unet.jl
```diff
@@ -0,0 +1,73 @@
...
+	decoder = Chain([Chain([
+		concat_layer[i],
+		decoder_layer[5-i]])
+					 for i in 4:-1:1]...)
+
+	final_conv = Conv((1, 1), features => out_channels, σ)
+
+	decoder = Chain(decoder, final_conv)
+
+	return layers
```
pri1311, line 48: Ok figured out, you are returning layers. Which does not have decoder and final_conv chained to it.

test/convnets.jl
theabhirath: What I did was simply run the formatter only in the `src` directory, and exclude everything else. It's a hack for now but it does most of the job

src/convnets/unet.jl
theabhirath: Agree with this, it looks weird – I think the only place we might need named layers is if we specifically need to index into a `Chain` later for use and the name of the layer isn't apparent. Here I don't see that happening.

src/convnets/unet.jl
```diff
@@ -0,0 +1,73 @@
...
+	features::Integer = 32)
+	encoder_conv = []
+	push!(encoder_conv,
+		unet_block(in_channels, features))
+
+	append!(encoder_conv,
+		[unet_block(features * 2^i, features * 2^(i + 1)) for i in 0:2])
+
+	encoder_conv = Chain(encoder_conv)
+	encoder_pool = [Chain(encoder_conv[i], MaxPool((2, 2); stride = (2, 2))) for i in 1:4]
```
theabhirath, line 22: I think you can combine the `MaxPool` and the `unet_block` functions into a separate block? That should help writing code with less of the indexing.

theabhirath:
Just saw this, and it looks good! I have some comments on the code style and implementation (I've already added some, and I will add some more later). But before that, I want to understand which meaning of the "U-Net" architecture is being used here and which could be a good fit for Metalhead. There are two ways in which this is interpreted in various implementations:

1. As it is in the original paper (and this is what @shivance seems to have done here). A good PyTorch example is https://github.com/milesial/Pytorch-UNet. I think this is a fairly standard model and, thus, would be good to have in the repository, but I have hardly ever seen it being used without some modifications. This brings me to point 2, which is
2. As a generic architecture with a CNN backbone for an encoder, where the upsampling encoder remains unchanged (save for some added flexibility). This reduces to 1 if used with a specific CNN backbone (the user always has the option of writing independently). https://gist.github.com/rwightman/f8b24f4e6f5504aba03e999e02460d31 is a gist I found from the author of `timm`, Ross Wightman, which explores this option.

Since Metalhead is moving towards a more flexible structure, I think the second option makes a lot of sense. We could also even write the default encoder from the paper.

shivance:
https://github.com/FluxML/Metalhead.jl/pull/210#discussion_r1058559895

Hi @theabhirath, thanks for the review. Even I was thinking of a flexible version of it.
But as of now, we have a blocker, to make `Parallel` work, as @darsnack suggested [here](https://github.com/FluxML/Metalhead.jl/pull/210#discussion_r1058559895).

Could you review that as well?

ToucheSir:
To be clear, using `Parallel` is not mandatory. I think the main goal of that comment was to decompose as much of that custom, monolithic (i.e. not modular) forward pass into existing Flux and Metalhead layers/functions. As inspiration, you can see what FastAI does [here](https://github.com/FluxML/FastAI.jl/blob/3ae742a1bbaf7d6d969b4de098ef05efa4d3d62e/FastVision/src/models/unet.jl#L2). There was actually an earlier attempt at porting that exact model to Metalhead (https://github.com/FluxML/Metalhead.jl/pull/137), but the PR was never followed through.

darsnack:
Just to prevent anyone from waiting on me: I agree with @ToucheSir's comment. #137 which is based on FastAI's implementation uses recurrence to construct the model with `Parallel`. I would recommend trying to bring #137 "merge-ready" in this PR.

src/utilities.jl
```diff
@@ -4,13 +4,13 @@ seconddimmean(x) = dropdims(mean(x; dims = 2); dims = 2)
 # utility function for making sure that all layers have a channel size divisible by 8
 # used by MobileNet variants
 function _round_channels(channels::Number, divisor::Integer = 8, min_value::Integer = 0)
-    new_channels = max(min_value, floor(Int, channels + divisor / 2) ÷ divisor * divisor)
-    # Make sure that round down does not go down by more than 10%
-    return new_channels < 0.9 * channels ? new_channels + divisor : new_channels
+	new_channels = max(min_value, floor(Int, channels + divisor / 2) ÷ divisor * divisor)
+	# Make sure that round down does not go down by more than 10%
+	return new_channels < 0.9 * channels ? new_channels + divisor : new_channels
```
CarloLucibello, line 9: all the indentation changes in this file are bad

shivance:
Great, Thanks @ToucheSir and @theabhirath for the reference!

Adopted their implementation with the following noticeable changes:
1. I've reused Metalhead's `basic_conv_bn` layer instead of convlayer from fastai
2. reused Metalhead's `basicblock` resnetblock instead of incorporating FastAI's ResBlock in Metalhead
3. Added a utility function `iterlayers` (Can it be avoided again ?)

PR is ready for review. 

Lastly a question, where to add usage example ?

shivance:
@CarloLucibello my bad ! There is an juliaformatter extension in my viscose which messed it, fixed it back

src/convnets/unet.jl
```diff
@@ -0,0 +1,97 @@
...
+function UNetBlock(m_child, inplanes, midplanes, outplanes = 2inplanes)
+    return Chain(;
+                 upsample = SkipConnection(Chain(; child = m_child,
+                                                 upsample = PixelShuffleICNR(midplanes,
+                                                                             midplanes)),
+                                           Parallel(cat_channels, identity,
+                                                    BatchNorm(inplanes))),
+                 act = xs -> relu.(xs),
+                 combine = UNetCombineLayer(inplanes + midplanes, outplanes))
+end
```
theabhirath, line 57: I must confess, I still do not see the need or the use for the `NamedTuple` layer syntax. Also, activation functions now work directly inside `Chain`s, so there's no need to write the explicit broadcast.

```suggestion
function UNetBlock(m_child, inplanes, midplanes, outplanes = 2 * inplanes)
    return Chain(SkipConnection(Chain(m_child, PixelShuffleICNR(midplanes, midplanes)),
                                Parallel(cat_channels, identity, BatchNorm(inplanes))),
                 relu, UNetCombineLayer(inplanes + midplanes, outplanes))
end
```

theabhirath (changes requested):
This seems to be shaping up really well! I will still need to have a closer look at the implementation, but I've posted some initial comments. One note, it would be nice to have a test to make sure that the model is behaving as expected. As for usage docs, you can add a section in the documentation (it's admittedly a little rough around the edges, but that is a work in progress).

src/convnets/unet.jl
```diff
@@ -0,0 +1,97 @@
...
+
+"""
+    UNet(backbone, inputsize)
+    DenseNet(transition_configs::NTuple{N,Integer})
+
+Creates a UNet model with specified backbone. Backbone of Any Metalhead model can be used as
+encoder .
+Set `pretrain = true` to load the model with pre-trained weights for ImageNet.
```
pri1311, line 65:
```suggestion

```

shivance:
Added documentation, unit tests ( let's see how CI turns out ), formatting.
Incorporated previous suggestions.

src/convnets/unet.jl
```diff
@@ -0,0 +1,124 @@
...
+"""
+struct UNet
+	layers::Any
+end
+@functor UNet
+
+function UNet(backbone::Vector{Any}; pretrain::Bool = false, inputsize::NTuple{4, Integer},
+	outchannels::Integer = 3)
+	layers = unet(backbone, inputsize, outchannels)
+    if pretrain
```
pri1311, line 118: inconsistent indentation 
```suggestion
    layers = unet(backbone, inputsize, outchannels)
    if pretrain
```

src/convnets/unet.jl
```diff
@@ -0,0 +1,124 @@
+function PixelShuffleICNR(inplanes, outplanes; r = 2)
+	return Chain(basic_conv_bn((1, 1), inplanes, outplanes * (r^2)),
```
pri1311, line 2:
```suggestion
    return Chain(basic_conv_bn((1, 1), inplanes, outplanes * (r^2)),
```
Metalhead uses a single tab only. You might need to configure JuliaFormatter to follow the same.

shivance:
Thanks! I'll take a look at formatting once Again in the morning.

test/convnets.jl
```diff
@@ -347,3 +347,11 @@ end
         _gc()
     end
 end
+
+@testset "UNet" begin
+    encoder = Metalhead.backbone(ResNet(18))
+    model = UNet(encoder; inputsize = (256, 256, 3, 1), outplanes = 10)
```
pri1311, line 353:
```suggestion
    model = UNet(encoder; inputsize = (256, 256, 3, 1), outchannels = 10)
```

shivance:
@ToucheSir Is there a guide which can help me test locally. Simply including runtests.jl raises errors like UNet not defined. I uninstalled preinstalled metalhead then it raises some conflict in metalhead imports.

Additionally, I'm not sure if Juliaformatter is working correctly at my end, as I format everytime before pushing, yet I've received comments continuously on formatting's end.

pri1311:
> Simply including runtests.jl raises errors like UNet not defined.

Could you show the exact error? Cause I tried running the test locally, I got a similar error, but it seemed to be caused by the wrong type of arguments passed. If thats the case, you might need to recheck the function definition.

src/utilities.jl
```diff
@@ -79,3 +79,6 @@ linear_scheduler(drop_prob::Nothing; depth::Integer) = fill(drop_prob, depth)
 function _checkconfig(config, configs)
     @assert config in configs "Invalid configuration. Must be one of $(sort(collect(configs)))."
 end
+
+# Utility function to return Iterator over layers, adopted from FastAI.jl
+iterlayers(m::Chain) = Iterators.flatten(iterlayers(l) for l in m.layers)
```
pri1311, line 84:
```suggestion
iterlayers(m::Chain) = Iterators.flatten(iterlayers(l) for l in m.layers)
iterlayers(m) = (m,)
```
Missing this maybe?

pri1311:
Errors for me.
```
 Error: layer SkipConnection(Chain(Chain(Parallel(PartialFunctions.PartialFunction{typeof(Metalhead.addact), Tuple{typeof(relu)}, NamedTuple{(), Tuple{}}}(Metalhead.addact, (NNlib.relu,), NamedTuple()), Chain(Conv((1, 1), 256 => 512, stride=2, bias=false), BatchNorm(512)), Chain(Conv((3, 3), 256 => 512, pad=1, stride=2, bias=false), BatchNorm(512), relu, Conv((3, 3), 512 => 512, pad=1, bias=false), BatchNorm(512))), Chain(Parallel(PartialFunctions.PartialFunction{typeof(Metalhead.addact), Tuple{typeof(relu)}, NamedTuple{(), Tuple{}}}(Metalhead.addact, (NNlib.relu,), NamedTuple()), identity, Chain(Conv((3, 3), 512 => 512, pad=1, bias=false), BatchNorm(512), relu, Conv((3, 3), 512 => 512, pad=1, bias=false), BatchNorm(512))), identity)), Chain(Any[Conv((1, 1), 512 => 2048, bias=false), BatchNorm(2048, relu)], PixelShuffle(2))), Parallel(cat_channels, identity, BatchNorm(256))), index 2 in Chain, gave an error with input of size (16, 16, 256, 1)
└ @ Flux ~/.julia/packages/Flux/kq9Et/src/outputsize.jl:107
UNet: Error During Test at /Users/priyanagda/Documents/opensource/Metalhead.jl/test/convnets.jl:352
  Got exception outside of a @test
  MethodError: objects of type Vector{Any} are not callable
  Use square brackets [] for indexing an Array.
```

shivance:
@ToucheSir CI is 🟢 for UNet

src/utilities.jl
```diff
@@ -79,3 +79,7 @@ linear_scheduler(drop_prob::Nothing; depth::Integer) = fill(drop_prob, depth)
...
+iterlayers(m) = (m,)
```
darsnack, line 85: Call this `flatten_chains` instead because it only flattens `Chain`.

src/convnets/unet.jl
```diff
@@ -0,0 +1,123 @@
...
+    end
+end
+
+function UNetBlock(m_child, inplanes, midplanes, outplanes = 2 * inplanes)
+    return Chain(SkipConnection(Chain(m_child,
+                                      PixelShuffleICNR(midplanes, midplanes)),
+                                Parallel(cat_channels, identity, BatchNorm(inplanes))),
+                 xs -> relu.(xs),
+                 UNetCombineLayer(inplanes + midplanes, outplanes))
+end
```
darsnack, line 54: Also another lower case function name

src/convnets/unet.jl
```diff
@@ -0,0 +1,123 @@
...
+                         inplanes, midplanes, outplanes)
+    end
+end
+
+function UNetBlock(m_child, inplanes, midplanes, outplanes = 2 * inplanes)
+    return Chain(SkipConnection(Chain(m_child,
+                                      PixelShuffleICNR(midplanes, midplanes)),
+                                Parallel(cat_channels, identity, BatchNorm(inplanes))),
+                 xs -> relu.(xs),
+                 UNetCombineLayer(inplanes + midplanes, outplanes))
```
darsnack, line 53:
```suggestion
    return Chain(SkipConnection(Chain(m_child,
                                      PixelShuffleICNR(midplanes, midplanes)),
                                Parallel(cat_channels, identity, BatchNorm(inplanes))),
                 relu,
                 UNetCombineLayer(inplanes + midplanes, outplanes))
```

darsnack (changes requested):
I was able to play with the implementation in a GH code space, and it appears to be implemented correctly. There are only some minor changes suggested.

The main thing remaining before merging is updating the docstrings. Specifically, I think we need to explain a bit better how a ResNet backbone is "reversed" in the decoder.

src/convnets/unet.jl
```diff
@@ -0,0 +1,123 @@
...
+    `UNet` does not currently support pretrained weights.
+
+See also [`Metalhead.UNet`](@ref).
+"""
+struct UNet
+    layers::Any
+end
+@functor UNet
+
+function UNet(backbone, inputsize::NTuple{4, Integer}, outplanes::Integer = 3;
```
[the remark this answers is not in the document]
theabhirath: Most of the original CNN series (read: VGG, ResNet, Inception, DenseNet) are used very commonly, but that choice should be on the user since the backbone is often domain dependent (not sure why, but I've seen more DenseNets for medical use cases and more Inceptions for autonomous vehicles and the like). So having the default backbone match the paper would be best

src/convnets/unet.jl
```diff
@@ -0,0 +1,123 @@
...
+Creates a UNet model with specified backbone. Backbone of Any Metalhead model can be used as
+encoder.
+([reference](https://arxiv.org/abs/1505.04597)).
+
+# Arguments
+
+    - `backbone`: The backbone layers to be used in encoder. 
+    	For example, `Metalhead.backbone(Metalhead.ResNet(18))` can be passed to instantiate a UNet with layers of
+    	resnet18 as encoder.
+    - `inputsize`: size of input image
```
pri1311, line 99: it could be helpful to mention the order of BCHW for input size.

shivance:
Still working on it, not ready for review yet

shivance:
This PR is ready for review.

.github/workflows/CI.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           - 'r"/*/SEResNet*"'
           - '[r"Res2Net", r"Res2NeXt"]'
           - '"Inception"'
-          - '"DenseNet"'
+          - '["UNet", "DenseNet"]'
```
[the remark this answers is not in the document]
darsnack: Bump

test/convnets.jl
darsnack, line 21: I think you ran the formatter on the full test suite again. Can you only run it on `src` like @theabhirath suggested?

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+                    final::Any = unet_final_block, fdownscale::Integer = 0, kwargs...)
+    backbonelayers = collect(flatten_chains(backbone))
+    layers = unetlayers(backbonelayers, imgdims; m_middle = unet_middle_block,
+                        skip_upscale = fdownscale, kwargs...)
+
+    outsz = Flux.outputsize(layers, imgdims)
+    layers = Chain(layers, final(outsz[end - 1], outplanes))
+
+    return layers
+end
```
darsnack, line 84: Can you use the same argument names here as in `UNet`. Also, this function can just be called `unet` to match the other models in the repo.

The `kwargs` should be documented explicitly.

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+        outplanes = isnothing(outplanes) ? inplanes : outplanes
+
+        return unet_block(Chain(layer, childunet),
+                          inplanes, midplanes, outplanes)
+    end
+end
+
+"""
+    build_unet(backbone, imgdims, outplanes::Integer,
+    final::Any = unet_final_block, fdownscale::Integer = 0, kwargs...)
```
darsnack, line 58: The indenting is off here

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+
+Creates a UNet model with specified backbone. Backbone of Any Metalhead model
+can be used as encoder.
+([reference](https://arxiv.org/abs/1505.04597)).
```
darsnack, line 62:
```suggestion
Creates a UNet model with specified convolutional backbone. 
Backbone of any Metalhead ResNet-like model can be used as encoder 
([reference](https://arxiv.org/abs/1505.04597)).
```

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+
+    outsz = Flux.outputsize(layers, imgdims)
+    layers = Chain(layers, final(outsz[end - 1], outplanes))
+
+    return layers
+end
+
+"""
+UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
+backbone = Metalhead.backbone(DenseNet(121)); pretrain::Bool = false)
```
darsnack, line 88: Indenting is off

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+
+Creates a UNet model with an encoder built of specified backbone. By default it uses DenseNet Backbone, however any  of
+Metalhead model can be used for encoder backbone
+([reference](https://arxiv.org/abs/1505.04597)).
```
darsnack, line 92:
```suggestion
Creates a UNet model with an encoder built of specified backbone. 
By default it uses [`DenseNet`](@ref) backbone, but any ResNet-like Metalhead model can be used for the encoder
([reference](https://arxiv.org/abs/1505.04597)).
```

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+
+See also [`Metalhead.UNet`](@ref).
+"""
+struct UNet
+    layers::Any
+end
+@functor UNet
+
+function UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
+              backbone = Metalhead.backbone(DenseNet(121)); pretrain::Bool = false)
```
darsnack, line 116: `backbone` should be a keyword?

src/utilities.jl
```diff
@@ -79,3 +79,19 @@ linear_scheduler(drop_prob::Nothing; depth::Integer) = fill(drop_prob, depth)
...
+"""
+The flatten_chains function takes in a single argument m, which can be of `Any` type.
+If the input m is of type `Chain`, the function returns an iterator that recursively flattens 
+the layers of the input Chain object and all its sub-layers. 
+It does this by calling the flatten_chains function on each element of m.layers and 
+passing it into the Iterators.flatten function, which flatten the layers into a single iterator.
+If the input m is not of type Chain, the function returns a tuple containing the single input m.
+This function can be useful when you want to traverse nested layers of a Chain object and flatten 
+them into a single iterator.
+"""
```
darsnack, line 94: This should be a complete docstring

darsnack (changes requested):

shivance:
It's funny how many rounds of reviews, architecture changes, this PR has had. Over a month since it's being reviewed 😆

shivance:
Thanks @ToucheSir !

Are we still going with n dimensional unet?

shivance:
.
> * Write proper devdocs for Flux and Metalhead. This is a much bigger project and almost certainly will be an ongoing one.

@ToucheSir Come to think of it, this could be a potential GSoD project!

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+    `UNet` does not currently support pretrained weights.
+
+See also [`Metalhead.UNet`](@ref).
+"""
+struct UNet
+    layers::Any
+end
+@functor UNet
+
+function UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
```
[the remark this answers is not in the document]
ToucheSir: Same as with 2D. Spatial dimensions x channels/features x batch size, so all but the last two assuming the usual memory layout.

darsnack:
Unfortunately, I think GSoCs are not allowed to be solely for documentation (I'll have to double check this). But you can propose it for GSoD!

shivance:
@ToucheSir @darsnack I'm willing to open a follow up PR to add N spatial dimensional support.
Let's get the PR for 2 dimension merged in first !
(It's kind of demotivating for me to drag this PR further after so many rounds, feels like no result of all this work) 😅

Open for review in current state...

src/convnets/unet.jl
```diff
@@ -0,0 +1,125 @@
...
+    - `encoder_backbone`: The backbone layers of specified model to be used as encoder. 
+    	For example, `Metalhead.backbone(Metalhead.ResNet(18))` can be passed to instantiate a UNet with layers of
+    	resnet18 as encoder.
+    - `pretrain`: Whether to load the pre-trained weights for ImageNet
+
+!!! warning
+
+    `UNet` does not currently support pretrained weights.
+
+See also [`Metalhead.UNet`](@ref).
```
darsnack, line 108:
```suggestion

See also [`Metalhead.unet`](@ref).
```

src/convnets/unet.jl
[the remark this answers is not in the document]
darsnack: @shivance I think the point is that you don't need any changes other than dropping the type restriction to generalize to more dimensions.

But we'd want to have that in the test, so we can save it for another PR if you'd like.

darsnack (changes requested):
This looks like it is ready to merge modulo one small docstring issue.

PRs can take time (sometimes extenuated by our ability to review frequently). In our case, this is even more true since FluxML is very community-driven. This makes our development extremely distributed, and it is important that PRs are "release ready" before merging. Otherwise, simple changes can get bottlenecked from release due to larger changes that require refactoring/polishing.

Your patience and hard work is very much appreciated. The long review time is not a reflection of your work, just a consequence of the fact that we're all contributing on a volunteer basis. Please don't feel discouraged! Some of the most prolific Julia contributors have high impact PRs that take months to get right. So you're in good company!

shivance:
@darsnack So I'll leave the signature of imsize as

`imsize::Dims{2} = (256, 256)`

for now?

Or make it

`UNet(imsize::Dims= (256, 256)`

darsnack (approved):
Let's leave it as is for now. Good job!

shivance:
Thanks @darsnack !

shivance:
Thank you @ToucheSir @darsnack !

PR #214, opened 2023-02-01
Mark resnet acctests broken

ToucheSir:
These failing CI makes it harder to spot other failures, especially in downstream CI. Just a band-aid fix until we can get new weights generated.


ToucheSir:
Ok, I think the last of the failures has been patched over and this is good to go. Do we want a separate issue to track regenerating weights for ResNet, WideResNet and ResNeXt?

darsnack:
Line 128 of the `convents.jl` tests has `depth` as defined. Can you delete that variable?

ToucheSir:
Also changed 139 since it looked similar, is that what you meant?

ToucheSir:
Looks like that did the trick

darsnack (approved):

PR #225, opened 2023-04-23
add pytorch/torchvision porting and comparison script

CarloLucibello:


theabhirath:
This script and an older version that @darsnack had posted as a gist both work, but I find that `loadmodel!` is taking an insane amount of memory for even small models like VGG-11 or ResNet-18. However I have only tried this on my PC and cannot be sure...can someone else confirm if this is an issue?

darsnack:
[Here](https://gist.github.com/darsnack/bfb8594cf5fdc702bdacb66586f518ef) is also a script for porting weights.

theabhirath:
> [Here](https://gist.github.com/darsnack/bfb8594cf5fdc702bdacb66586f518ef) is also a script for porting weights.

Yep, that's the one I mentioned. Here too `loadmodel!` seems to eat up a ton of memory.

ToucheSir:
Can you take a quick allocation profile with and show the flamegraph here using PProf or the VSCode profile view?

theabhirath:
Sure. Got this with PProf. It's a screenshot of the flame graph, but if you want the raw results file I can try and upload that as well to somewhere else (I don't think GitHub supports sharing that).

CleanShot 2023-04-24 at 08 10 25

theabhirath:
Ah, @ToucheSir, I neglected to mention something which on a re-run I find might be the cause of this. I was using @darsnack's script to port weights but I added an `fmapstructure(identity, weights)` in the end before saving because IIRC we had discussed this would avoid saving information like the RNGs and allow us to save and load weights independent of Julia version (as would be ideal). When I remove that line, the memory requirement goes down to something manageable again. Is `loadmodel!` somehow very bad for loading models stored in this format? Can this be fixed?

PR #226, opened 2023-04-23
Revert "remove some duplicate include"

ToucheSir:
Reverts FluxML/Metalhead.jl#223 per https://github.com/FluxML/Metalhead.jl/pull/223#issuecomment-1519092766.

Changes to re-introduce:
- Updated gitignore
- rm Publish.jl config in Project.toml
- rm duplicate includes
- `dropblock([rng = default_rng_value(x)], ...` -> `dropblock([rng], ...`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,6 @@
 *.jl.*.cov
 *.jl.mem
 *.bson
-.vscode/
-.DS_Store
 
 # manifests
 docs/Manifest.toml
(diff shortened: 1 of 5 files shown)


CarloLucibello:
what is the change that doesn't  make sense? can we revert only that?

ToucheSir:
Everything not in the above list. I'm fine with either way, but figured this would be faster since there are other PRs almost ready to merge.

PR #229, opened 2023-04-24
fix the torchvision porting script

CarloLucibello:
The models left uncommented in the `model_list` in the `port_torchvision.jl` script are successfully ported from torchvision
(on julia 1.9 rc2). Is there any need to reupload them on HuggingFace?

For SqueezeNet and DenseNet the script errors due to size mismatches. Maybe wrong ordering again?

close #228

PR #230, opened 2023-04-25
fix ViT model output + rewrite attention layer + adapt torchvision script

### LINKED ISSUE #231: cannot match attention layer output to pytorch's one
I'm struggling to import the weights from torchvision's ViT to ours. The problem is that the correct map of the attention layers in torch to the one in metalhead seems non-trivial. 

```julia
using PythonCall, Metalhead
torch = pyimport("torch")

function th2jl(x::Py)
    xj = pyconvert(Array, x.detach().numpy())
    xj = permutedims(xj, ndims(xj):-1:1)
    return xj
end

m = torch.nn.MultiheadAttention(embed_dim=2, num_heads=1, batch_first=true, bias=false, add_bias_kv=false)
# python forward pass
x = torch.randn(1, 3, 2)
y, a = m(x, x, x, need_weights=true)

mj = Metalhead.MHAttention(2, 1, qkv_bias=false)
# copy weights
mj.qkv_layer.weight .=  th2jl(m.in_proj_weight)'    # transpose back since Linear layers in pytorch don't need transpose 
mj.projection.layers[1].weight .=  th2jl(m.out_proj.weight)' 
# julia forward pass
xj = th2jl(x)
yj = mj(xj)

@assert yj ≈ th2jl(y) # false
```
Probably this is due to the permutations and chunking in our initial projection, possibly we should rearrange them in such a way that the natural weight mapping from pytorch just works.

Pinging @theabhirath for more insights.

CarloLucibello:
Current status of this PR is that all weights are copied but the outputs on the test image differ (and flux's one don't make sense)

close #231

### DIFF
--- a/src/layers/Layers.jl
+++ b/src/layers/Layers.jl
@@ -37,7 +37,7 @@ include("classifier.jl")
 export create_classifier
 
 include("normalise.jl")
-export prenorm, ChannelLayerNorm
+export prenorm, ChannelLayerNorm, LayerNormV2
 
 include("pool.jl")
 export AdaptiveMeanMaxPool
--- a/src/layers/Layers.jl
+++ b/src/layers/Layers.jl
@@ -37,7 +37,7 @@ include("classifier.jl")
 export create_classifier
 
 include("normalise.jl")
-export prenorm, ChannelLayerNorm, LayerNormV2
+export prenorm, ChannelLayerNorm
 
 include("pool.jl")
 export AdaptiveMeanMaxPool
--- a/src/layers/Layers.jl
+++ b/src/layers/Layers.jl
@@ -16,7 +16,7 @@ import Flux.testmode!
 include("../utilities.jl")
 
 include("attention.jl")
-export MHAttention
+export MultiHeadSelfAttention
 
 include("conv.jl")
 export conv_norm, basic_conv_bn, dwsep_conv_norm
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -1,5 +1,5 @@
 # Utility function for applying LayerNorm before a block
-prenorm(planes, fn) = Chain(LayerNorm(planes), fn)
+prenorm(planes, fn) = Chain(LayerNormV2(planes), fn)
 
 """
     ChannelLayerNorm(sz::Integer, λ = identity; ϵ = 1.0f-6)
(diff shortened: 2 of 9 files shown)


theabhirath:
I'm not sure that I want the porting scripts to be generalised to all models. We used it for CNNs because it's convenient but even there the script is not exactly always directly usable (for example, SqueezeNets require you to remove the `reverse` added in #229 for the script to work. At best we perhaps link the scripts in the docs somewhere but having it in the repo IMO seems to suggest to users that somehow Metalhead cannot be used by them to train models on their own. The ideal solution, of course, is to train all the models in Metalhead on Imagenet and host those weights. But since there are certain obstacles to that which I encountered last summer, porting only the CNN weights (and leaving the porting script as a link instead of directly in the repo) is a good middle ground I feel. However, this is only my opinion. @darsnack and @ToucheSir may also think of something?

ToucheSir:
I agree we need not be worried about having a single one-size-fits-all script. If one script per model family/group of model families helps simplify the porting code, that sounds good to me.

CarloLucibello:
The problem with ViT is in the attention module, probably the weights have to be copied in some particular fashion, I will have to investigate further. 

I really want to get ViT in because it is the most popular vision backbone these days.

CarloLucibello:
After rewriting the attention layer on top of NNlib and removing the final tanh from ViT I can reproduce pytorch's outputs although there is still a slight discrepancy:
```julia
Flux:
    acoustic guitar: 0.90519154
    stage: 0.0040107034
    harmonica: 0.0028614246
    microphone: 0.002621256
    electric guitar: 0.0025401094
PyTorch:
    acoustic guitar: 0.90745604
    stage: 0.0038461224
    harmonica: 0.002782756
    microphone: 0.0025289422
    electric guitar: 0.0023941135
```
This could be due to differences in the implementation of layer norm see https://github.com/FluxML/Flux.jl/pull/2220

CarloLucibello:
I'm happy with this. If i can get an approval I'll merge and move on

CarloLucibello:
Looks like changing the implementation of LayerNorm has little effect:
```
Flux (LayerNormV2):
    acoustic guitar: 0.9051971
    stage: 0.0040056095
    harmonica: 0.0028621724
    microphone: 0.0026183864
    electric guitar: 0.0025359727
Flux:
    acoustic guitar: 0.90519154
    stage: 0.0040107034
    harmonica: 0.0028614246
    microphone: 0.002621256
    electric guitar: 0.0025401094
PyTorch:
    acoustic guitar: 0.90745604
    stage: 0.0038461224
    harmonica: 0.002782756
    microphone: 0.0025289422
    electric guitar: 0.0023941135
```
so I don't know why we observe these discrepancies. I added LayerNormv2 to the Layers module but didn't use it anywhere since I'm not sure it will be really needed

src/layers/Layers.jl
```diff
@@ -37,7 +37,7 @@ include("classifier.jl")
 export create_classifier
 
 include("normalise.jl")
-export prenorm, ChannelLayerNorm
+export prenorm, ChannelLayerNorm, LayerNormV2
```
darsnack, line 40: Let's not export this yet.

src/layers/normalise.jl
```diff
@@ -1,5 +1,5 @@
 # Utility function for applying LayerNorm before a block
-prenorm(planes, fn) = Chain(LayerNorm(planes), fn)
+prenorm(planes, fn) = Chain(LayerNormV2(planes), fn)
```
darsnack, line 2:
```suggestion
prenorm(planes, fn) = Chain(LayerNorm(planes), fn)
```

src/vit-based/vit.jl
```diff
@@ -100,9 +102,10 @@ end
 @functor ViT
 
 function ViT(config::Symbol; imsize::Dims{2} = (256, 256), patch_size::Dims{2} = (16, 16),
-             pretrain::Bool = false, inchannels::Integer = 3, nclasses::Integer = 1000)
+             pretrain::Bool = false, inchannels::Integer = 3, nclasses::Integer = 1000, 
+             qkv_bias=false)
```
darsnack, line 106: Unless it is typical to adjust this toggle, I think it should not get exposed going from `vit` to `ViT`. The logic with the codebase has been to make the uppercase exports as simple as possible.

darsnack (approved):
A few minor changes before merging but otherwise looks good

src/vit-based/vit.jl
CarloLucibello: I add to add it since the default for torchvision is `true`, here was `false`. The torchvision model is given by
```

src/layers/normalise.jl
CarloLucibello: good catch

src/vit-based/vit.jl
darsnack: Okay, so change the default to `true` and remove the keyword? I assume you almost always want it as `true`.

CarloLucibello: yes, I'll do it in the next PR

PR #233, opened 2023-05-05
automatize huggingface upload + change ViT defaults

### LINKED ISSUE #232: change ViT constructor's defaults
See https://github.com/FluxML/Metalhead.jl/pull/230#discussion_r1186068260

CarloLucibello:
Fix #232 

Also updates the torchvision conversion script to serialize with JLD2 the state from `Flux.state` instead of the full model. 

I have already uploaded the model states to huggingface
https://huggingface.co/FluxML


darsnack:
I don't the HF uploads are in the right format for Artifacts to work. Steps 3+ in the [contributing guide](https://github.com/FluxML/Metalhead.jl/blob/master/docs/src/contributing.md) need to be followed to create the tarball, get the SHAs/URLs, and update the Artifacts.toml in this PR.

Otherwise looks good.

CarloLucibello:
added a script automatizing the huggingface uploads and regenerated Artifacts.toml (preserving also the old artifacts for the time being)

darsnack (approved):

PR #241, opened 2023-05-25
RFC: DenseNet rewrite for correctness

theabhirath:
Our previous version of `DenseNet` does not combine features at the block level, which is one of the defining innovations of the paper. This one manages to do that, but at the cost of looking a little clunky (it has an explicitly written out `DenseBlock` and uses `vcat` for vector concatenation). It also does not work currently on Zygote 0.6.45+ due to https://github.com/FluxML/Zygote.jl/issues/1417. Any comments on how to make the design work with less "manual" and more built-in layers are appreciated! This version may also not be very AD-friendly, so any help there is also very much appreciated 😅.

cc @ToucheSir @darsnack


src/convnets/densenet.jl
```diff
@@ -1,20 +1,32 @@
...
-                                conv_norm((3, 3), expansion * outplanes, outplanes;
-                                          pad = 1, revnorm = true)...), cat_channels)
+function (m::DenseBlock)(x)
+    input = [x]
+    for layer in m.layers
+        x = layer(input)
+        input = vcat(input, [x])
+    end
+    return cat_channels(input...)
 end
```
theabhirath, line 30: This is where I think the biggest change is, and where I think the code could use the most input. Is this the best way of doing this sequence of operations?

src/convnets/densenet.jl
```diff
@@ -1,20 +1,32 @@
...
+                            pad = 1, revnorm = true)...,
+                 Dropout(dropout_prob))
+end
 
-Create a Densenet bottleneck layer
-([reference](https://arxiv.org/abs/1608.06993)).
+function dense_block(inplanes::Integer, num_layers::Integer, bn_size::Integer,
+                     growth_rate::Integer, dropout_prob)
+    layers = [dense_bottleneck(inplanes + (i - 1) * growth_rate, growth_rate, bn_size,
+                                   dropout_prob) for i in 1:num_layers]
```
darsnack, line 14: Can't you take this vector and build the nested `Parallel` of a `DenseBlock` recursively? That's how I would try doing it to avoid defining an extra type.

darsnack (changes requested):

src/convnets/densenet.jl
theabhirath: I'm afraid I can't quite immediately see that implementation 😅 Could you give me an example?

src/convnets/densenet.jl
```diff
@@ -1,20 +1,32 @@
-"""
-    dense_bottleneck(inplanes, outplanes; expansion=4)
+function dense_bottleneck(inplanes::Integer, growth_rate::Integer, bn_size::Integer,
+                          dropout_prob)
+    return Chain(cat_channels,
```
CarloLucibello, line 3: If the input to this chain is a vector of arrays, according tohttps://github.com/FluxML/Metalhead.jl/blob/010d4bc72989e4392d2ce0b89e72b8d640927dd8/src/utilities.jl#L37
shouldn't we have
```suggestion
    return Chain(x -> cat_channels(x...),
```
?

src/convnets/densenet.jl
CarloLucibello: maybe 
```julia
function (m::DenseBlock)(x)
    input = (x,)
    for layer in m.layers
        x = layer(input)
        input = (input..., x)
    end
    return cat_channels(input)
end
```
to bypass the Zygote `vcat` problem?

CarloLucibello: Even better, we can remove `cat_channels` from the layers and do 
```julia
function (m::DenseBlock)(x)
    input = x
    for layer in m.layers
        x = layer(input)
        input = cat_channels(input, x)
    end
    return input
end

**Review by CarloLucibello — APPROVED** [review]

**Review by CarloLucibello — APPROVED** [review]

**Review by CarloLucibello — APPROVED** [review]

**Review by darsnack — CHANGES_REQUESTED** [review]
Unless I am missing something, I don't see how the old implementation is functionally different (modulo dropout). I drew a picture. Let's not merge this until we make sure.

image

**Comment by darsnack** [comment]
The actual dense net paper has this weird way of drawing the operation that suggest that we want some recursive structure. But concatenation along an axis is inherently cumulative, so that accumulation is implicit?

**Comment by darsnack** [comment]
And this also seems the same as [torchvision](https://github.com/pytorch/vision/blob/d2f7486ccaef461913cdb51990ff353addf6f064/torchvision/models/densenet.py#L96-L124). So I guess @theabhirath can you roll back the changes introducing `DenseBlock` and just keep the smaller pieces like adding dropout?

**Comment by CarloLucibello** [comment]
but still there should be something different from torchvision since the two forwards don't match

**Comment by CarloLucibello** [comment]
The paper says
```julia
x1 = H1([x0])
x2 = H1([x0, x1])
x3 = H1([x0, x1, x2])
...
```
which is equivalent to 
```julia
input = [x0]
x = H1(input)
input = [input..., x]
x = H2(input)
input = [input..., x]
```
Torchvision does the same. So it seems to me that 1) the implementation in this PR is equivalent to the old one with skip connection; 2) it is also equivalent to the 2 variants I proposed above; 3) they are all equivalent to the one in torchvision, and the difference in the forwards is not caused concat mechanism but it's somewhere else.

darsnack:
Yeah agree with Carlo here.

> Notice that in this image, the way the outputs of the layers (and the inputs) are treated is that they are produced, remembered and concatenated at the end.

That's because this is a poorly drawn figure. They are re-showing the concatenation of every group of channels into each bottleneck. This is because the output of the concatenation (where arrows merge) goes into the `i`-th bottleneck block but nowhere else. So they have to redraw that concatenation + the output of the `i`-block going into the `i + 1`-th block. They could just as easily have drawn an arrow from where the arrows merge into the `i`-th block to the `i + 1`-th block (which would have been the same diagram with fewer lines, and a closer match to how you would programmatically implement this sequence of blocks).

theabhirath:
Okay, I spent a lot of time poring over the paper and figured out that it is indeed what we were doing 😅Apologies for the noise. This is what I get for trying to map Pythonic programming styles onto Julia 😅. However, maybe I can keep this open until I figure out where it is that we differ from the torchvision implementation. Given that the we are doing the correct thing, though, this should not block v0.8 (we can release without DenseNet weights, I think). I will land a final docs PR over this weekend and then we can release that.

PR #248, opened 2023-07-16
Remove NNlibCUDA dependency to support Flux 0.14

christiangnrd:
This is in regards to the question I posted on the slack. I asked if NNlibCUDA was a necessary dependency for Metalhead since there is no direct reference to it in the code and it's preventing Flux 0.14 from being used in the same environment as Metalhead.

I made this a draft because I don't know the full repercussions of this change. I'm seeking feedback from more experienced Metalhead/Flux developers.


CarloLucibello:
Seems fine. Could CUDA as well be removed?

christiangnrd:
There are 2 `dropblock_mask` functions for cuda in `src/layers/drop.jl`.

I played around with putting them in an extension but there are design decisions that I am not comfortable making so I figured this would be a good solution for until a proper extension can be made.

darsnack (approved):

PR #249, opened 2023-07-22
Adding a more detailed introduction to the `Layers` module

theabhirath:
This PR adds a more detailed introduction to the `Layers` module than there previously was. This is the first in a long series of documentation upgrades that I hope to land slowly over the next few months, thoroughly documenting the depth (pun intended 😁) Metalhead.jl offers post v0.8. I slowly hope to convert all the plain API reference pages into walkthroughs that link to examples or just directly show them on the page itself, so that usage and API live in the same place where possible.

I would love comments on the general flow of documentation we should be aspiring for as well as pertaining to this particular case since this is the first PR and so it should hopefully set a good precedent moving forward 😁

cc @CarloLucibello @darsnack @ToucheSir


theabhirath:
Bump on this, any comments?

src/convnets/builders/invresmodel.jl
```diff
@@ -9,16 +9,6 @@
...
-
-# Arguments
-
-    - `scalings`: a tuple of two numbers that specify the width and depth scaling factors.
-    - `block_configs`: This is a `Vector` of `Tuple`s that specifies the configuration of the
-    inverted residual blocks. This can take several forms:
-        + for `dwsep_conv_norm`, or depthwise separable convolutional blocks, the tuple
-        should be of the form `(dwsep_conv_norm, kernel size, output channels, stride,
-        number of repeats, activation function)`. For example, the following configuration
-        is valid: `(dwsep_conv_norm, 3, 64, 1, 1, relu6)`.
```
ToucheSir, line 21: Why was this part of the docstring removed? Otherwise LGTM and we can always do small cleanups to post-merge since this documentation has been wanted for a while :)

theabhirath: I wanted to add more detailed docs in a follow-up PR, this was incomplete and thus sometimes misleading 😅

ToucheSir (approved):

PR #254, opened 2023-10-16
Update use of `Flux._isactive` in DropBlock to prevent deprecation warning

theabhirath:
In the process of moving Dropout to NNlib, some of these functions were changed upstream. While this is obviously not ideal for Metalhead to use internal Flux functions, this is a temporary fix to prevent a deprecation warning from being printed every time the layer is used. Maybe later Flux can find a way to do this upstream that allows other libraries to extend.

cc @ToucheSir @darsnack to check if this is the way to go for now.


darsnack (approved):
Great! Can you also bump the version to 0.9 so we can release?

darsnack (approved):

PR #258, opened 2023-11-03
Adding Shuffle

RafaelT00:
[Please delete this text and describe your change here.
For bugfixes, please detail the bug and include a test case which your patch fixes.
If you are adding a new feature, please clearly describe the design, its rationale, the possible alternatives considered.
It is easiest to merge new features when there is clear precedent in other systems; we need to know we're taking
the right direction since it can be hard to change later.]

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/convnets/shufflenet.jl
+++ b/src/convnets/shufflenet.jl
@@ -153,8 +153,8 @@ function shufflenet(groups, width_scale, num_classes; in_channels=3)
         channels,
         init_block_channels,
         groups; 
-        in_channels=in_channels, 
-        num_classes=num_classes)
+        in_channels, 
+        num_classes)
 
     return net
 end
--- a/src/convnets/shufflenet.jl
+++ b/src/convnets/shufflenet.jl
@@ -1,0 +1,160 @@
+using Flux
+
+"""
+Channelshuffle(channels, groups)
+
+Channel shuffle operation from 'ShuffleNet: An Extremely Efficient Convolutional Neural Network for Mobile Devices
+([reference](https://arxiv.org/abs/1707.01083)).
+
+# Arguments
+
+  - `channels`: number of channels
+  - `groups`: number of groups
+"""
+function ChannelShuffle(x::Array{Float32, 4}, g::Int)
+    width, height, channels, batch = size(x)
+    channels_per_group = channels÷g
+    if (channels % g) == 0
+        x = reshape(x, (width, height, g, channels_per_group, batch))
+        x = permutedims(x,(1,2,4,3,5))
+        x = reshape(x, (width, height, channels, batch))
+    end
+    return x
+end
+
+"""
+ShuffleUnit(in_channels::Integer, out_channels::Integer, grps::Integer, downsample::Bool, ignore_group::Bool)
+
+Shuffle Unit from 'ShuffleNet: An Extremely Efficient Convolutional Neural Network for Mobile Devices
+([reference](https://arxiv.org/abs/1707.01083)).
+
+# Arguments
+
+  - `in_channels`: number of input channels
+  - `out_channels`: number of output channels
+  - `groups`: number of groups
+  - `downsample`: apply downsaple if true
+  - `ignore_group`: ignore group convolution if true
+"""
+function ShuffleUnit(in_channels::Integer, out_channels::Integer, groups::Integer, downsample::Bool, ignore_group::Bool)
+    mid_channels = out_channels ÷ 4
+    groups = ignore_group ? 1 : groups
+    strd = downsample ? 2 : 1
+
+    if downsample
+        out_channels -= in_channels
+    end
+
+    m = Chain(Conv((1,1), in_channels => mid_channels; groups,pad=SamePad()),
+              BatchNorm(mid_channels),
+              NNlib.relu,
+              x -> ChannelShuffle(x, groups),
+              DepthwiseConv((3,3),  mid_channels => mid_channels; bias=false, stride=strd, pad=SamePad()),
+              BatchNorm(mid_channels),
+              NNlib.relu,
+              Conv((1,1), mid_channels => out_channels; groups, pad=SamePad()),
+              BatchNorm(out_channels),
+              NNlib.relu)
+    
+    if downsample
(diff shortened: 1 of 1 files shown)


src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+
+# Arguments
+
+  - `channels`: number of channels
+  - `groups`: number of groups
+"""
+function ChannelShuffle(x::Array{Float32, 4}, g::Int)
+    width, height, channels, batch = size(x)
+    channels_per_group = channels÷g
+    if (channels % g) == 0
```
ToucheSir, line 17:
```suggestion
    if channels % g == 0
```
We have a JuliaFormatter config in this repo, so make sure to run that before pushing your code.

src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+
+Channel shuffle operation from 'ShuffleNet: An Extremely Efficient Convolutional Neural Network for Mobile Devices
+([reference](https://arxiv.org/abs/1707.01083)).
+
+# Arguments
+
+  - `channels`: number of channels
+  - `groups`: number of groups
+"""
+function ChannelShuffle(x::Array{Float32, 4}, g::Int)
```
ToucheSir, line 14:
```suggestion
function channel_shuffle(x::AbstractArray{Float32, 4}, g::Int)
```
This type constraint is too restrictive. If `ChannelShuffle` works for all number types than it should reflect that. Generally all utility functions in Metalhead need to be GPU-compatible too. The renaming is a suggestion for how to make this function more "Julian", since it's not a callable type (which would be PascalCase) but a plain function. Lastly, how does this handle 3D inputs?

src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+    groups = ignore_group ? 1 : groups
+    strd = downsample ? 2 : 1
+
+    if downsample
+        out_channels -= in_channels
+    end
+
+    m = Chain(Conv((1,1), in_channels => mid_channels; groups,pad=SamePad()),
+              BatchNorm(mid_channels),
+              NNlib.relu,
```
ToucheSir, line 50:
```suggestion
              BatchNorm(mid_channels, relu),
```
`relu` is already in scope because of `using NNlib` and fusing it into the preceeding norm is slightly more efficient. Also, is the activation function not configurable for ShuffleNet?

src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+              x -> ChannelShuffle(x, groups),
```
ToucheSir, line 51:
```suggestion
              Base.Fix2(channel_shuffle, groups),
```
Will be easier on the compiler.

src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+              DepthwiseConv((3,3),  mid_channels => mid_channels; bias=false, stride=strd, pad=SamePad()),
+              BatchNorm(mid_channels),
+              NNlib.relu,
+              Conv((1,1), mid_channels => out_channels; groups, pad=SamePad()),
+              BatchNorm(out_channels),
+              NNlib.relu)
+    
+    if downsample
+        m = Parallel((mx, x) -> cat(mx, x, dims=3),m, MeanPool((3,3); pad=SamePad(), stride=2))
```
ToucheSir, line 60:
```suggestion
        m = Parallel(cat_channels, m, MeanPool((3,3); pad=SamePad(), stride=2))
```
We have [`cat_channels`](https://github.com/FluxML/Metalhead.jl/blob/d60172f54c1ede0aa68f3eeb5eb84f822a3373c4/src/utilities.jl#L31) for this exact case.

src/convnets/shufflenet.jl
```diff
@@ -0,0 +1,160 @@
...
+            out_ch::Integer = trunc(out_channels)
+            push!(stage, ShuffleUnit(in_channels, out_ch, groups, downsample, ignore_group))
+            in_channels = out_ch
+        end
+        append!(features, stage)
+    end
+  
+    model = Chain(features...)
+  
+    return Chain(model, GlobalMeanPool(), Flux.flatten, Dense(in_channels => num_classes))
```
ToucheSir, line 105:
```suggestion
    return Chain(model, GlobalMeanPool(), MLUtils.flatten, Dense(in_channels => num_classes))
```
The general modus operandi of this library has been to create named types for the top-level model and wrap the underlying Chain with them. You can see this pattern in the files for any of the other exported models.

For the suggestion. `flatten` is only imported and not defined in Flux. It's preferable to use a symbol from the library that actually defined when that library is available (which MLUtils is, being a dep of Metalhead).

ToucheSir (commented):
Thanks for the contribution! This is a great start, and the next steps would be adding tests + better matching the code style of the rest of the repo.

RafaelT00:
I made the suggested changes

src/convnets/shufflenet.jl
RafaelT00: I didn't think about it when writing the function, so for a 3D inputs, a batch of grey images, would be necessary to artificially create a channel dimension.

src/convnets/shufflenet.jl
RafaelT00: Applied.

src/convnets/shufflenet.jl
RafaelT00: Changed.

src/convnets/shufflenet.jl
RafaelT00: Changed.

src/convnets/shufflenet.jl
RafaelT00: Could I see an example? Sorry, I'm still a newbie using Julia, I looked to the rest of convnets and tried to code with a similar style, but there are still things I that still haven't fully understood.

ToucheSir:
Thanks for the updates. On a quick skim nothing stands out to me, can you add it to the test suite to finish off the PR?

PR #259, opened 2023-12-04
Allow loading custom weights files for EfficientNet

IanButterworth:
Intends to make it possible to load custom pretrained EfficientNet models.

Opening this PR as a draft to see if there are opinions as I am unfamiliar with this package.

### PR Checklist

- [ ] Tests are added


ToucheSir:
This generally makes sense to me, can you think of any way taking a string path could limit us further down the line @darsnack @theabhirath ?

darsnack:
There was a plan to open up the `pretrain` argument to allow specifying the name of an artifact to load. The intention being to have multiple artifacts per model spec.

But I think the way this is implemented, it should be easy to add that case down the line. Looks good to me!

PR #260, opened 2023-12-04
use julia-actions/cache in CI

IanButterworth:
Just a fly-by suggestion while creating #259 

julia-actions/cache is now more powerful and stores precompile caches

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,11 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):
Nice to see it working now. How might we observe the performance improvements from caching? Also, is there an equivalent for Buildkite (to help solve our current timeout issues)?

IanButterworth:
If you can, I recommend moving to the standard `julia-actions/julia-runtest` because it sets up coverage in a faster way and does parallel precompilation. The current setup here is hitting old serial precompilation (which is silently happening at the start of the test run)

IanButterworth:
Multiplexing test groups via an env var that's handled in `runtests.jl` should be doable?

ToucheSir:
Another problem: 
image
Trying to normalize these suite names would be more yaml-foo than I want to deal with right now (because we still need them for ReTest to work properly). Is there no way to make have the cache action work with them, given that they are still valid paths on *nix systems?

IanButterworth:
That should be fixed upstream in the cache action

PR #266, opened 2023-12-26
Move to Documenter 1.0 (and some miscellaneous fixes)

theabhirath:
Tidied up our Documenter workflow a little and also migrated to Documenter 1.0 (which has some more features and has the added benefit of looking nicer!).

One thing that may have been confusing here is the `utilities.jl` file. It's been listed as part of the `Layers` module on the docs but hasn't been in that folder? So that's been fixed. The file is still sourced in Metalhead proper though.

Some other miscellaneous fixes:

1. Not sure what the proper way is to refer to `Flux.Conv` with a doc reference, but whatever we were doing before wasn't working. Removed the link for now.
2. Tidied up some of the docstrings with references to stop throwing warnings.
3. Added a `using Downloads` to stop a deprecation warning from popping up every time we test.

This is probably a start for some PRs for the docs, hopefully I can find time and improve them further soon.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -113,14 +113,6 @@ jobs:
             using Pkg
             Pkg.develop(PackageSpec(path=pwd()))
             Pkg.instantiate()'
-      - run: |
-          julia --color=yes --project=docs/ -e '
-            using Metalhead
-            # using Pkg; Pkg.activate("docs")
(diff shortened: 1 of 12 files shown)


ToucheSir (approved):
Those Buildkite errors are unexpected, but I also don't see how any of the changes here could be causing them.

PR #270, opened 2024-01-04
Deprecate older VGG API

### LINKED ISSUE #269: VGG has two "higher-level" functions
VGG currently has two functions in the "higher-level" API category (as per the camel-casing nomenclature):

```julia
VGG(imsize::Dims{2}; config, inchannels, batchnorm = false, nclasses, fcsize, dropout_prob)
VGG(depth::Integer; pretrain::Bool = false, batchnorm::Bool = false, inchannels::Integer = 3, nclasses::Integer = 1000)
```

The former is probably an artifact of pre-0.8 Metalhead, since there is also a mid-level function for VGG with a very similar API:

```julia
vgg(imsize; config, inchannels, batchnorm = false, nclasses, fcsize, dropout_prob)
```

This came up in #268, where the pre-trained version of VGG has its docstring hidden. We can work around this I believe by writing a single docstring with both variants, but if it is agreeable maybe we can deprecate the first function above for uniformity?

theabhirath:
Fix #269. Also adds the pre-trained API to the docstring of VGG. And a small rewrite of the VGG block, which was written pre-`conv_norm`'s current more powerful form.

Right now the `depwarn` isn't printed by default, should we force it to be true? Or is it reasonable to assume that people will spot this and not upgrade if they don't want to/migrate to the newer API if they want to?


theabhirath:
P.S. there has to be a way to limit the tests run when we open a PR, because running ResNet/MobileNet tests for a VGG PR seems hugely wasteful. Is there a hook to GH Actions that can let us specify which tests to run via a comment or something?

src/convnets/vgg.jl
```diff
@@ -110,10 +104,37 @@ const VGG_CONV_CONFIGS = Dict(:A => [(64, 1), (128, 1), (256, 2), (512, 2), (512
...
+        inchannels::Integer = 3, nclasses::Integer = 1000)
 
 Construct a VGG model with the specified input image size. Typically, the image size is `(224, 224)`.
 
+!!! warning
+
+    The `VGG(imsize; config, inchannels, batchnorm, nclasses)` function documented below
+    will be deprecated in a future release. Please use `vgg(imsize; config, inchannels, 
+    batchnorm, nclasses)` instead for the same functionality.
+
```
darsnack, line 137: I would not document a deprecated method

src/convnets/vgg.jl
```diff
@@ -174,3 +169,18 @@ function VGG(depth::Integer; pretrain::Bool = false, batchnorm::Bool = false,
...
+# Deprecated; to be removed in a future release
+function VGG(imsize::Dims{2}; config, batchnorm::Bool = false, dropout_prob = 0.5,
+    inchannels::Integer = 3, nclasses::Integer = 1000)
+    Base.depwarn("The `VGG(imsize; config, inchannels, batchnorm, nclasses)` constructor 
+                will be deprecated in a future release. Please use `vgg(imsize; config, 
+                inchannels, batchnorm, nclasses)` instead for the same functionality.", :VGG)
+    layers = vgg(imsize; config, inchannels, batchnorm, nclasses, dropout_prob)
+    return VGG(layers)
+end
```
darsnack, line 186: Can we move deprecations to a `deprecations.jl` file?

darsnack (changes requested):

darsnack:
I disagree—passing tests always seem wasteful 😁. IMO tests that don't always run are as bad as no tests at all.

theabhirath:
> I disagree—passing tests always seem wasteful 😁. IMO tests that don't always run are as bad as no tests at all.

That makes sense 😂 I was thinking more from the GH actions perspective of how many free minutes we get but I think the individual test sets are executing pretty quickly now so it should be okay

darsnack:
Public repos have unlimited minutes, so that's not an issue. I guess the only thing is that a single org shares runners across all repos, and there are a fixed amount of runners you can use in parallel. But I haven't noticed this to be an issue; it's not like we are waiting significantly longer to merge PRs.

theabhirath:
> Public repos have unlimited minutes, so that's not an issue.

Oh, for some reason I thought this was limited 😅 That's perfect then, waiting a bit for a PR to merge is not a problem at all.

darsnack (approved):

PR #272, opened 2024-01-15
Fix VGG dropout probability

### LINKED ISSUE #271: Default VGG dropout probability changed to zero
PR #270 inadvertently changed the default `Dropout` probability of VGG models from `0.5` to `0.0`:

```julia
julia> VGG(11)
VGG(
  Chain(
    Chain(
      Conv((3, 3), 3 => 64, relu, pad=1),  # 1_792 parameters
      MaxPool((2, 2)),
      Conv((3, 3), 64 => 128, relu, pad=1),  # 73_856 parameters
      MaxPool((2, 2)),
      Conv((3, 3), 128 => 256, relu, pad=1),  # 295_168 parameters
      Conv((3, 3), 256 => 256, relu, pad=1),  # 590_080 parameters
      MaxPool((2, 2)),
      Conv((3, 3), 256 => 512, relu, pad=1),  # 1_180_160 parameters
      Conv((3, 3), 512 => 512, relu, pad=1),  # 2_359_808 parameters
      MaxPool((2, 2)),
      Conv((3, 3), 512 => 512, relu, pad=1),  # 2_359_808 parameters
      Conv((3, 3), 512 => 512, relu, pad=1),  # 2_359_808 parameters
      MaxPool((2, 2)),
    ),
    Chain(
      MLUtils.flatten,
      Dense(25088 => 4096, relu),       # 102_764_544 parameters
      Dropout(0.0),                                                    <==
      Dense(4096 => 4096, relu),        # 16_781_312 parameters
      Dropout(0.0),                                                    <==
      Dense(4096 => 1000),              # 4_097_000 parameters
    ),
  ),
)                   # Total: 22 arrays, 132_863_336 parameters, 506.839 MiB.
```

There currently also isn't any direct way to create a pretrained VGG model with custom `dropout_prob`: 
```Julia-repl
julia> VGG(11; pretrain=true, dropout_prob=0.5)
ERROR: MethodError: no method matching VGG(::Int64; pretrain::Bool, dropout_prob::Float64)
```

adrhill:
Closes #271 by setting default dropout probability to 0.5 and making keyword argument `dropout_prob` accessible.

### DIFF
--- a/src/convnets/vgg.jl
+++ b/src/convnets/vgg.jl
@@ -103,7 +103,7 @@ const VGG_CONFIGS = Dict(11 => [(64, 1), (128, 1), (256, 2), (512, 2), (512, 2)]
 
 """
     VGG(depth::Integer; pretrain::Bool = false, batchnorm::Bool = false,
-        inchannels::Integer = 3, nclasses::Integer = 1000, dropout_prob = 0.5)
+        inchannels::Integer = 3, nclasses::Integer = 1000)
 
 Create a VGG style model with specified `depth`.
 ([reference](https://arxiv.org/abs/1409.1556v6)).
--- a/src/convnets/vgg.jl
+++ b/src/convnets/vgg.jl
@@ -103,13 +103,13 @@ const VGG_CONFIGS = Dict(11 => [(64, 1), (128, 1), (256, 2), (512, 2), (512, 2)]
 
 """
     VGG(depth::Integer; pretrain::Bool = false, batchnorm::Bool = false,
-        inchannels::Integer = 3, nclasses::Integer = 1000)
+        inchannels::Integer = 3, nclasses::Integer = 1000, dropout_prob = 0.5)
 
 Create a VGG style model with specified `depth`.
 ([reference](https://arxiv.org/abs/1409.1556v6)).
 
 !!! warning
-    
+
     `VGG` does not currently support pretrained weights for the `batchnorm = true` option.
 
 # Arguments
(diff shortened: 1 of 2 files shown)


src/convnets/vgg.jl
```diff
@@ -128,9 +129,9 @@ end
 @functor VGG
 
 function VGG(depth::Integer; pretrain::Bool = false, batchnorm::Bool = false,
-             inchannels::Integer = 3, nclasses::Integer = 1000)
+             inchannels::Integer = 3, nclasses::Integer = 1000, dropout_prob = 0.5)
     _checkconfig(depth, keys(VGG_CONFIGS))
-    layers = vgg((224, 224); config = VGG_CONFIGS[depth], batchnorm, inchannels, nclasses)
+    layers = vgg((224, 224); config = VGG_CONFIGS[depth], batchnorm, inchannels, nclasses, dropout_prob)
```
theabhirath, line 134: Not sure if we should introduce `dropout_prob` for the `VGG` function. Aside from the fact that this would be breaking, Metalhead has the `vgg` function for exactly this sort of customisation. What do you think @darsnack?

theabhirath (changes requested):

src/convnets/vgg.jl
theabhirath: Xref https://github.com/FluxML/Metalhead.jl/issues/271#issuecomment-1892604680 for a way to load weights without this, but maybe this is worth it if only to avoid that extra code?

darsnack: It's not breaking as long as the default probability matches pre-#270. But I agree that it's better to leave it off for API consistency.

adrhill: Addressed in [69b34c3](https://github.com/FluxML/Metalhead.jl/pull/272/commits/69b34c3dc386ae63b3e4efa1f791518e2bfe8618) and [be19ebf](https://github.com/FluxML/Metalhead.jl/pull/272/commits/be19ebf6054c9f73ff74e7a30a7f4b66d28e2e86).

theabhirath (approved):
Awesome, thanks for the fix! Could you also bump the version so that we can release?

Issue #243, opened 2023-06-06
UNet inchannels != 3 not working


nickkeepfer:
It seems the UNet implementation only works with inputs with channel-size 3
```
Status `/private/var/folders/98/8xz00xq10hq7kdrdvdjd1jy00000gn/T/jl_EZ324U/Project.toml`
  [dbeba491] Metalhead v0.8.0
  [44cfe95a] Pkg v1.8.0
```
```julia
using Metalhead
UNet((128,128),1,3,Metalhead.backbone(DenseNet(121)))
```
`ERROR: DimensionMismatch: layer Conv((7, 7), 3 => 64, pad=3, stride=2, bias=false) expects size(input, 3) == 3, but got 128×128×1×1 Array{Flux.NilNumber.Nil, 4}`

theabhirath:
Hi @nickkeepfer, as I understand it, the way U-Net is being constructed right now requires you to specify `inchannels` at two places - one for the function itself and another for the backbone. So `UNet((128,128),1,3,Metalhead.backbone(DenseNet(121; inchannels = 1)))` should hopefully yield the model that you want. Do let me know if this works the way you expect it to. If it does, there can be a discussion around reducing this redundancy.

theabhirath:
Yeah, it's definitely weird to keep it this way, I will try and land a PR soon to address this.

PR #276, opened 2024-03-22
Fix UNet implementation with arbitrary channel sizes (#243)

vinayakjeet:
#243 

Bug Description:
The current UNet implementation in the Metalhead package has a limitation where it only works with input tensors of channel size 3. This restriction causes compatibility issues when users try to use UNet with input tensors of different channel sizes.

Patch Description:
To address this limitation, I've modified the UNet implementation to support input tensors with arbitrary channel sizes. The UNet model can now handle input with varying dimensions

Test Case:
using Metalhead
UNet((128,128),1,3,Metalhead.backbone(DenseNet(121)))

This UNet model can process without any errors

### DIFF
--- a/src/convnets/unet.jl
+++ b/src/convnets/unet.jl
@@ -71,18 +71,21 @@ Backbone of any Metalhead ResNet-like model can be used as encoder
     - `final`: final block as described in original paper
     - `fdownscale`: downscale factor
 """
-function unet(encoder_backbone, imgdims, outplanes::Integer,
-              final::Any = unet_final_block, fdownscale::Integer = 0)
-    backbonelayers = collect(flatten_chains(encoder_backbone))
-    layers = unetlayers(backbonelayers, imgdims; m_middle = unet_middle_block,
-                        skip_upscale = fdownscale)
+function unet(encoder_backbone, imgdims, inchannels::Integer, outplanes::Integer,
+    final::Any = unet_final_block, fdownscale::Integer = 0)
+backbonelayers = collect(flatten_chains(encoder_backbone))
 
-    outsz = Flux.outputsize(layers, imgdims)
-    layers = Chain(layers, final(outsz[end - 1], outplanes))
+# Adjusting input size to include channels
+adjusted_imgdims = (imgdims..., inchannels, 1)
 
-    return layers
+layers = unetlayers(backbonelayers, adjusted_imgdims; m_middle = unet_middle_block,
+              skip_upscale = fdownscale)
+
+outsz = Flux.outputsize(layers, adjusted_imgdims)
+layers = Chain(layers, final(outsz[end - 1], outplanes))
+
+return layers
 end
-
 """
     UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
          encoder_backbone = Metalhead.backbone(DenseNet(121)); pretrain::Bool = false)
@@ -114,7 +117,7 @@ end
 
 function UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
               encoder_backbone = Metalhead.backbone(DenseNet(121)); pretrain::Bool = false)
-    layers = unet(encoder_backbone, (imsize..., inchannels, 1), outplanes)
+    layers = unet(encoder_backbone, imsize, inchannels, outplanes)
     model = UNet(layers)
     if pretrain
         artifact_name = "UNet"


theabhirath:
Hi Vinayakjeet, thanks for the PR! Unfortunately, I don't think this does what we want yet. The problem is that `inchannels` isn't being passed to the model backbone. What you've done is try and change the input being passed in to the `Flux.outputsize` function, which actually causes an error when I try to initialise the model:

```julia
julia> using Metalhead

julia> model = UNet((128,128),1,3,Metalhead.backbone(DenseNet(121)))
ERROR: DimensionMismatch: layer Conv((7, 7), 3 => 64, pad=3, stride=2, bias=false) expects size(input, 3) == 3, but got 128×128×1×1 Array{Flux.NilNumber.Nil, 4}
Stacktrace:
  [1] _size_check(layer::Flux.Conv{2, 2, typeof(identity), Array{…}, Bool}, x::Array{Flux.NilNumber.Nil, 4}, ::Pair{Int64, Int64})
    @ Flux ~/.julia/packages/Flux/jgpVj/src/layers/basic.jl:195
  [2] (::Flux.Conv{2, 2, typeof(identity), Array{Float32, 4}, Bool})(x::Array{Flux.NilNumber.Nil, 4})
    @ Flux ~/.julia/packages/Flux/jgpVj/src/layers/conv.jl:198
  [3] #outputsize#340
    @ ~/.julia/packages/Flux/jgpVj/src/outputsize.jl:93 [inlined]
  [4] outputsize(m::Flux.Conv{2, 2, typeof(identity), Array{Float32, 4}, Bool}, inputsizes::NTuple{4, Int64})
    @ Flux ~/.julia/packages/Flux/jgpVj/src/outputsize.jl:91
  [5] unetlayers(layers::Vector{…}, sz::NTuple{…}; outplanes::Nothing, skip_upscale::Int64, m_middle::typeof(Metalhead.unet_middle_block))
    @ Metalhead ~/Code/Metalhead.jl/src/convnets/unet.jl:34
  [6] unet(encoder_backbone::Flux.Chain{…}, imgdims::Tuple{…}, inchannels::Int64, outplanes::Int64, final::typeof(Metalhead.unet_final_block), fdownscale::Int64)
    @ Metalhead ~/Code/Metalhead.jl/src/convnets/unet.jl:81
  [7] unet
    @ ~/Code/Metalhead.jl/src/convnets/unet.jl:76 [inlined]
  [8] #UNet#175
    @ ~/Code/Metalhead.jl/src/convnets/unet.jl:120 [inlined]
  [9] UNet(imsize::Tuple{Int64, Int64}, inchannels::Int64, outplanes::Int64, encoder_backbone::Flux.Chain{Tuple{…}})
    @ Metalhead ~/Code/Metalhead.jl/src/convnets/unet.jl:118
 [10] top-level scope
    @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

I would suggest that you try and rewrite the function in such a way that `inchannels` is passed along to the encoder backbone.

src/convnets/unet.jl
```diff
@@ -114,7 +117,7 @@ end
 
 function UNet(imsize::Dims{2} = (256, 256), inchannels::Integer = 3, outplanes::Integer = 3,
               encoder_backbone = Metalhead.backbone(DenseNet(121)); pretrain::Bool = false)
-    layers = unet(encoder_backbone, (imsize..., inchannels, 1), outplanes)
+    layers = unet(encoder_backbone, imsize, inchannels, outplanes)
```
theabhirath, line 120: `inchannels` should somehow be passed in to the encoder backbone here. Of course, we will have to decide how to deal with this in case the user passes in a model with this initialised and also separately `inchannels`

vinayakjeet:
I've updated the PR with fixes for the DimensionMismatch error by adjusting the encoder's first convolutional layer to match input channels. Please review the changes and share your feedback.

src/convnets/unet.jl
```diff
@@ -71,18 +71,34 @@ Backbone of any Metalhead ResNet-like model can be used as encoder
...
     - `fdownscale`: downscale factor
 """
-function unet(encoder_backbone, imgdims, outplanes::Integer,
-              final::Any = unet_final_block, fdownscale::Integer = 0)
-    backbonelayers = collect(flatten_chains(encoder_backbone))
-    layers = unetlayers(backbonelayers, imgdims; m_middle = unet_middle_block,
-                        skip_upscale = fdownscale)
+function unet(encoder_backbone, imgdims, inchannels::Integer, outplanes::Integer,
+    final::Any = unet_final_block, fdownscale::Integer = 0)
+backbonelayers = collect(flatten_chains(encoder_backbone))
```
CarloLucibello, line 76: please pay attention to the formatting, you lost the indentation here

vinayakjeet:
A beginner contributor to the codebase, can you review the logic I have implemented, additionally I have encountered an error MethodError indicating a mismatch in method signatures for the unet function. It appears that there might be an issue with how the encoder_backbone is instantiated or utilized within the unet function. Could you please review the instantiation and usage of the encoder_backbone

